PR #7, opened 2022-11-14
Finish writing math.py with PyTorch

PROA200:
I modified the _math.py file from the TensorFlow M3GNet repository to use PyTorch and added it to megnet/utils/math.py.

Other changes:
- I added the tests from the other repository for math.py; this is essentially identical except for one location (tf/torch).
- config.py was slightly modified to use PyTorch datatypes.
- I made some small modifications to setup.py to keep the .npy data file during pip installation.

Elliott


shyuep:
This looks great. To solve the linting problems, you can pip install black and just run black on the megnet and tests folders. That will auto-reformat all the code.

PR #10, opened 2022-12-05
M3 g net setup

kenko911:
SiLU activation is added


shyuep:
Pls fix the linting issues and also merge the current main changes. The linting is due to indentation, which is quite serious for python. I suggest you use something like Pycharm to code (the free Community version will do). The IDE will take care of most of these issues.

PR #13, opened 2022-12-07
MEGNet Dataloader class is added

kenko911:


megnet/dataloader/MEGNetDataset.py
```diff
@@ -0,0 +1,134 @@
...
+        Check if the dgl_graph.bin exists or not
+        """
+        graph_path = str(self.path) + "/dgl_graph.bin"
+        return os.path.exists(graph_path)
+
+    def process(self):
+        """
+        Call _load_graph to convert Pymatgen structure into dgl graphs
+        """
+        self.graphs = self._load_graph()
```
shyuep, line 93: This is unnecessary? Why not just put the code from self._load_graph() in the process method?

megnet/dataloader/MEGNetDataset.py
```diff
@@ -0,0 +1,134 @@
...
+        num_graphs = self.labels.shape[0]
+        graphs = []
+
+        for idx in trange(num_graphs):
+            structure = self.structures[idx]
+            graph, state_attr = self.cry_graph.get_graph_from_structure(structure=structure)
+            graphs.append(graph)
+        return graphs
+
+    def save(self):
```
shyuep, line 108: I think you can provide a filename argument to allow people to specify the filename. The default can be `dgl_graph.bin`.

megnet/dataloader/MEGNetDataset.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Save dgl graphs
+        """
+        graph_path = str(self.path) + "/dgl_graph.bin"
+        labels_with_key = {self.label_name: self.labels}
+        save_graphs(graph_path, self.graphs, labels_with_key)
+
+    def load(self):
```
shyuep, line 116: Similarly, load can take a filename.

megnet/dataloader/MEGNetDataset.py
kenko911: Hi, Shyue Ping, I just tried to put the code in the process method and it doesn't work in the expected way meaning that the length of training, validation and testing dataloader are incorrect. I am trying to figure it out why but I believe there maybe some operations done in DGLDataset class.

megnet/dataloader/MEGNetDataset.py
kenko911: Yup, I can add that

megnet/dataloader/MEGNetDataset.py
kenko911: Yup, I can add that.

megnet/dataloader/MEGNetDataset.py
shyuep: I don't think this is very likely. The process method is a one-liner that sets the self.graph.

megnet/dataloader/MEGNetDataset.py
```diff
@@ -0,0 +1,134 @@
...
+        structures: Pymatgen strutcure
+        labels: property values
+        label: label name
+        cry_graph: Pmg2Graph
+        """
+        self.cry_graph = cry_graph
+        self.structures = structures
+        self.labels = torch.FloatTensor(labels)
+        self.label_name = label_name
+        self.path = os.getcwd()
```
shyuep, line 79: This is not necessary. All paths are by default from the current working directory. You do not need this variable.

PR #20, opened 2023-01-14
Multi gpu training

JiQi535:
Multiple GPU training of megnet with QM9 data.

Training 1000 epochs with 1, 2, 4 GPUs consumed 43.80 (100%), 22.85 (52%) and 12.51 (29%) hours. While training requires more epochs to converge with multiple GPUs, the convergence of train mse_loss is faster with more GPUs, and validation l1_loss also is converging faster with more GPUs.

Also, though training with 1 GPU converges the fastest with respect to training error, its validation loss turns out to be highest. I'm not sure about the reason, and maybe the randomness in training process is also critical, and we may need to train multiple models and select the best one from them. 

megnet-QM9

@shyuep The current train.py is still process oriented, and the functions in QM9_train.py are not implemented in the module. Do you have any suggestions on where to put those functions and where to have a trainer class? Thanks.

### DIFF
--- a/examples/qm9_utils.py
+++ b/examples/qm9_utils.py
@@ -89,16 +89,19 @@ def set_seed(seed: int) -> None:
     dgl_seed(seed)
 
 
-def create_dataloaders(config: Munch, data: tuple):
+def create_dataloaders(config: Munch, data: tuple, use_ddp=False):
     dataloaders = namedtuple("Dataloaders", ["train", "val", "test"])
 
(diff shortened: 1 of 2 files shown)


shyuep:
I think you can create a new module called training to implement helper classes for training with data. QM9 should be an example and a test file, but the implementation should work with any dataset. If needed, the crystal and molecule fitting can be separate classes/methods.

shyuep:
I assume this is now defunct.

PR #40, opened 2023-04-21
Pretrained MEGNet models for formation energy and band gap are added

kenko911:


shyuep:
This is great. But I would like you to add the actual training script in the examples. This is useful reference for people who want to train their own models.

PR #58, opened 2023-05-11
remove direct import from "private" module in example

lbluque:
This was bugging me.


shyuep:
Thanks. Missed it when I was refactoring.

PR #70, opened 2023-06-09
Package health

janosh:
@arosen93 Re https://github.com/materialsproject/pymatgen/issues/3051, I think the paths should be fine. Windows has become quite robust w.r.t. Unix style paths. I didn't see any version pins in `setup.py`. Could you specify? Happy to fix as part of this PR.

@shyuep You don't have a `MANIFEST.in` file so I think there's no point in having `include_package_data=True`?

https://github.com/materialsvirtuallab/matgl/blob/0c2dc3b9f08ffbad3bb66c00dadb0f0a0937ee48/setup.py#L47

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ ci:
 
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.270
+    rev: v0.0.272
     hooks:
       - id: ruff
         args: [--fix, --ignore, D]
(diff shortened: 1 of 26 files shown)


janosh:
Also, rather than putting the version in `matgl/__init__.py`, I would treat `setup.py` as the source of truth for this and then do sth like this in `matgl/__init__.py`.

```py
from importlib.metadata import PackageNotFoundError, version

try:
    __version__ = version("matgl")
except PackageNotFoundError:
    pass  # package not installed
```

arosen93:
Looks like the pins were either removed or I was looking at the wrong thing :) Will try `matgl` out on Windows another time and let you fix any path issues (one of my favorite PRs to make to most codes).

janosh:
@arosen93 Good catch. `pymatgen` is used all over the place. Even in the core package

https://github.com/materialsvirtuallab/matgl/blob/465a53bfcd31cd80d45eb71c9b37c4774ffa2c92/matgl/layers/_atom_ref.py#L9

so should definitely be a dep. Same is true for `ase` and `pytorch_lightning`.

PR #74, opened 2023-06-09
Stack lattice matrices in converters

lbluque:
Stack `lattice_matrices` to single `np.array` before creating the corresponding tensor. This prevents the following warning:

```
UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at  ../torch/csrc/utils/tensor_new.cpp:210.)
  return th.as_tensor(data, dtype=dtype)
```

Actually has a noticeable speed-up.


shyuep:
Many thanks.

PR #80, opened 2023-06-14
Refactor basis function classes

lbluque:
A proposal to refactor the basis function classes (ie spherical harmonics and Bessel functions to their own module), in order to allow cleaner addition of new basis function classes (ie Fourier basis, etc).


lbluque:
:wave:  Hi matgl team, thanks so much for the hard work!

I am looking to port CHGNet to matgl, and this seems like a clean way to add a few additional classes needed.

Please let me know any comments or suggestions with regard to this PR and I can update accordingly, and definitely feel free to directly push edits.

The things I have done so far:
- create a `basis` module
- move `SphericalHarmonics` to `basis._spherical.py`
- move `GaussianExpansion` and `SphericalBessel` to `basis._radial`
- move `combine_sbf_shf` to `layers._three_body`

I am unsure if a new module `basis` is the best design, perhaps it should go into `layers` either as a submodule or simply the files `spherical_.py`, `radial_.py` directly if you prefer a flat layout.

shyuep:
Many thanks! @kenko911 thoughts?

My own inclination and suggestions:
1. `_basis.py` in layers package.
2. Single file for the different functions (rather than separate one for radial vs spherical etc.). These are not long enough to warrant separation. We can do so in the future if it grows unweildly. 
3. There are some minor doc/code changes. Pls merge the main branch into your PR and fix any minor merge conflicts.

kenko911:
Hi @lbluque, thank you very much for your contributions to refactoring basis functions. I agree that we should have _basis.py since we can implement different basis functions for two, three or even four body in the future. I checked the code and it looks good to me and we can merge these changes. I just have a question regarding CHGNet. I guess you will have different three-body interactions and graph convolutions to be implemented in _three_body.py and  _graph_convolutions.py. Right?

kenko911:
Great! please let me know if any help is needed from our side.

lbluque:
@kenko911 I still need to refactor the unit-tests accordingly before this PR is ready. I have two couple of questions:

Is `spherical_bessel_smooth` in `utils.maths` used or planned to be used?

Also, have you considered writing tests with pytest? Not for this PR perse, but I find it quite a lot better than unit-test, so it may be worth moving over before tests become unwieldy

shyuep:
Pls use pytest for the unittests. I used pytest in the newer parts of the code. But too lazy to refactor the old parts. In any case, pytest detects both old and new types of tests.

lbluque:
> Pls use pytest for the unittests. I used pytest in the newer parts of the code. But too lazy to refactor the old parts. In any case, pytest detects both old and new types of tests.

Got it. I'll then write pytest tests for any additional functionality in the future!

kenko911:
@lbluque  I would prefer leaving spherical_bessel_smooth in _basis.py to allow user to choose either using smooth or non-smooth spherical Bessel functions for constring MLIPs or property models.

lbluque:
Ok, I'm just confused since the `spherical_bessel_smooth` is not called in `SphericalBesselFunction._call_smooth_rbf`. Actually after searching for usages `spherical_bessel_smooth` does not seem to be called at all in any other function, which is why I left it in `utils.maths`.

shyuep:
@lbluque You can move it to _basis too. It is not used at the moment but no harm leaving it in for now.

lbluque:
I left `SphericalBesselWithHarmonics` in layers._three_body.py, but can move it to basis_.py if that's better.

Otherwise, this is ready to merge, unless you have any further suggestions.

shyuep:
Thanks!. Merged. I will do the refactoring.

PR #87, opened 2023-06-21
Fourier expansion layer and generalize cutoff polynomial

lbluque:
- Implements a `FourierExpansion` basis layer.
- Generalized `polynomial_cutoff` to allow setting the min exponent as well. I left the default exponent = 3 to keep the behavior unchanged.
- Added/modified tests


matgl/utils/cutoff.py
```diff
@@ -7,17 +7,27 @@
...
+        exponent (int): minimum exponent of the polynomial. Default is 5.
+            The polynomial will include terms of order exponent, exponent + 1, exponent + 2.
 
-    Returns: polynomial cutoff functions
-
+    Returns: polynomial cutoff function
     """
+    a = -(exponent + 1) * (exponent + 2) / 2
+    b = exponent * (exponent + 2)
+    c = -exponent * (exponent + 1) / 2
```
janosh, line 26: Are these single-letter var names to match the paper? If not, maybe there are more descriptive names?

matgl/utils/cutoff.py
```diff
@@ -7,17 +7,27 @@
...
+    """Envelope polynomial function that ensures a smooth cutoff.
+
+    Ensures first and second derivative vanish at cuttoff. As described in:
+        https://arxiv.org/abs/2003.03123
+
     Args:
         r (torch.tensor): radius distance tensor
         cutoff (float): cutoff distance.
+        exponent (int): minimum exponent of the polynomial. Default is 5.
+            The polynomial will include terms of order exponent, exponent + 1, exponent + 2.
```
janosh, line 20: Default is 3

matgl/utils/cutoff.py
```diff
@@ -7,17 +7,27 @@
 import torch
 
 
-def polynomial_cutoff(r, cutoff: float):
-    """Polynomial cutoff function
+def polynomial_cutoff(r: torch.tensor, cutoff: float, exponent: int = 3) -> torch.tensor:
```
janosh, line 10:
```suggestion
def polynomial_cutoff(r: torch.Tensor, cutoff: float, exponent: int = 3) -> torch.Tensor:
```

matgl/utils/cutoff.py
lbluque: good catch! will fix

shyuep:
I suggest you merge the main branch first. I have already fixed many of the typing issues previously.

shyuep:
Thanks.

PR #93, opened 2023-06-24
Unit test of isolated atoms and two-body for MEGNet and M3GNet models are added

kenko911:


JiQi535:
Just want to confirm about one special case for training. Will this implementation work for a training set of all isolated atoms or a training set of all structures without three-body interactions? @kenko911

JiQi535:
@kenko911 I just downloaded and tested the code. All unit tests are passed, but some simple calculations for energy of isolated atoms seems not working with bugs. Below is the code I tried to run but failed when lattice parameter ≥ 5.774 Å. 

```
import matgl
from pymatgen.core import Lattice, Structure

print(f"The atoms are isolated after lattice parameter increase to over {5/3**0.5*2} Å")
pot = matgl.load_model("M3GNet-MP-2021.2.8-PES")
lattice_parameters = [5.5, 5.773, 5.774, 6, 7, 8]
for lattice_parameter in lattice_parameters:
    struct_Iso = Structure.from_spacegroup("Pm-3m", Lattice.cubic(lattice_parameter), 
                                           ["Zr", "Zr"], [[0, 0, 0], [0.5, 0.5, 0.5]])
    atoms = AseAtomsAdaptor().get_atoms(struct_Iso)
    calc = M3GNetCalculator(pot)
    atoms.set_calculator(calc)
    print(f"The matgl calculated energy is {float(atoms.get_potential_energy()):.17f} eV when lattice parameter is {lattice_parameter} Å.")
```
Could you try and let me know if this is the case? Thanks.

matgl/ext/pymatgen.py
```diff
@@ -124,10 +164,14 @@ def get_graph(self, structure: Structure) -> tuple[dgl.DGLGraph, list]:
...
-        g.edata["lattice"] = tensor(np.stack([lattice_matrix for _ in range(g.num_edges())]))
+        # no neighbor
+        if len(src_id) == 0:
+            g = get_empty_graph(structure.num_sites)
+        else:
+            u, v = tensor(src_id), tensor(dst_id)
+            g = dgl.graph((u, v))
+            g.edata["pbc_offset"] = torch.tensor(images)
+            g.edata["lattice"] = tensor([[lattice_matrix] for i in range(g.num_edges())])
         g.ndata["attr"] = tensor(Z)
```
JiQi535, line 175: `Z` might have a larger size than the `g.num_nodes()` when `src_id` does not contain all site ids. For example, for a structure with three-body interactions in some regions but isolated atoms in other regions, the `src_id` does not contain the ids of those isolated atoms, so that the graph `g` will have less number of nodes than `len(Z)`. I think the code will fail for those cases, am I right?

JiQi535:
Also, I wanted to test a type of structure that I think this implementation would fail, i.e., structures with 3-body interactions only in some regions of the structure while isolated atoms in other regions. I think prediction of this type of structure might fail because the dimension of variable `Z` might be larger than the number of nodes of a graph `g`, which is defined by  `src_id`. (see related comment [here](https://github.com/materialsvirtuallab/matgl/pull/93/files#r1240916607)) I could not  test it now because of the above discussed issues when trying to run the code, but do you think this will be a problem?

kenko911:
> @kenko911 I just downloaded and tested the code. All unit tests are passed, but some simple calculations for energy of isolated atoms seems not working with bugs. Below is the code I tried to run but failed when lattice parameter ≥ 5.774 Å.
> 
> ```
> import matgl
> from pymatgen.core import Lattice, Structure
> 
> print(f"The atoms are isolated after lattice parameter increase to over {5/3**0.5*2} Å")
> pot = matgl.load_model("M3GNet-MP-2021.2.8-PES")
> lattice_parameters = [5.5, 5.773, 5.774, 6, 7, 8]
> for lattice_parameter in lattice_parameters:
>     struct_Iso = Structure.from_spacegroup("Pm-3m", Lattice.cubic(lattice_parameter), 
>                                            ["Zr", "Zr"], [[0, 0, 0], [0.5, 0.5, 0.5]])
>     atoms = AseAtomsAdaptor().get_atoms(struct_Iso)
>     calc = M3GNetCalculator(pot)
>     atoms.set_calculator(calc)
>     print(f"The matgl calculated energy is {float(atoms.get_potential_energy()):.17f} eV when lattice parameter is {lattice_parameter} Å.")
> ```
> 
> Could you try and let me know if this is the case? Thanks.

Hi Ji, thanks a lot for your comments, the problem comes from the Atoms2Graph class. The variable "state_attr" is assigned as a list object. We have to convert it into torch tensor.

JiQi535:
> Hi Ji, thanks a lot for your comments, the problem comes from the Atoms2Graph class. The variable "state_attr" is assigned as a list object. We have to convert it into torch tensor.

I have converted it into torch tensor, but other bugs keep coming out.

matgl/ext/pymatgen.py
kenko911: The g.ndata["attr"] is a one-hot vector of elements and the dimensionality of this tensor would be (num_atoms, len(element_list)), which should match the dimensionality of "Z" even they are isolated atoms.

kenko911:
Yup I think the graph.ndata["volume"] should also change the data type. Like this:g.ndata["volume"] = tensor([volume for i in range(atomic_number.shape[0])], dtype=torch.float32)

matgl/ext/pymatgen.py
JiQi535: Yes. The problem is that `g.num_nodes()` will be different with the dimensionality of `Z` for structures with a portion of atoms with 3-b interactions while the other atoms are isolated, and this will lead to a mismatch of `g.num_nodes()` with the dimensionality of node feature `g.ndata["attr"]`, which is not allowed for a dgl graph.

shyuep:
Pls fix all the merge conflicts before I merge.

kenko911:
> Pls fix all the merge conflicts before I merge.

Hi Shyue Ping, I think Ji already included my changes. I will close the merge request. Thanks!

PR #123, opened 2023-08-07
Extend the predict functionality for multi-fidelity MEGNet model using mgl command

kenko911:


shyuep:
Add a unit test.

matgl/cli.py
```diff
@@ -73,10 +74,27 @@ def predict_structure(args):
     """
     model = matgl.load_model(args.model)
     if args.infile:
-        for f in args.infile:
-            structure = Structure.from_file(f)
-            val = model.predict_structure(structure)
-            print(f"{args.model} prediction for {f}: {val}.")
+        if args.model == "MEGNet-MP-2019.4.1-BandGap-mfi":
+            state_dict = ["PBE", "GLLB-SC", "HSE", "SCAN"]
+            if len(args.infile) == 1 or not args.infile[1].isdigit():
```
shyuep, line 79: Counting args should not be the way this is implemented. The infile is supposed to take in multiple filenames. Not a filename and a state attr. State attr needs to be a separate, **optional** arg that is used only for mfi or models requiring a state_attr. E.g., `--state`.

kenko911: I agree. Let me change it.

PR #125, opened 2023-08-07
Add potential with additional site wise predictions

lbluque:
Since CHGNet also predicts site-wise (magmoms), we need a `Potential` that allows that.

### DIFF
--- a/matgl/apps/pes.py
+++ b/matgl/apps/pes.py
@@ -56,26 +56,11 @@ class Potential(nn.Module, IOMixIn):
         self.data_mean = data_mean if data_mean is not None else torch.zeros(1)
         self.data_std = data_std if data_std is not None else torch.ones(1)
 
-    def forward(
-        self, g: dgl.DGLGraph, state_attr: torch.Tensor | None = None, l_g: dgl.DGLGraph | None = None
-    ) -> tuple:
-        """Args:
-            g: DGL graph
-            state_attr: State attrs
-            l_g: Line graph.
-
-        Returns:
-            energies, forces, stresses, hessian: torch.Tensor
-        """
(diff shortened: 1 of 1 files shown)


lbluque:
This is a suggestion to minimize repeated code.

Please let me know what you think @kenko911  and @shyuep

shyuep:
Looks good to me. Though maybe we can just have one Potential class that includes a kwarg indicating if there is a site-wise prediction?

lbluque:
> Looks good to me. Though maybe we can just have one Potential class that includes a kwarg indicating if there is a site-wise prediction?

Ok, that sounds good and is even simpler!

shyuep:
Just a quick note that right now, the Calculator and Relaxer are called M3GNetCalculator. If there is no good reason why the calculator needs to be tied to a specific potential, we can rename it simply as MGLCalculator instead. Then the Calculator and Relaxer classes are agnostic to whether it is M3gnet, CHGNet or whatever Net that comes after.

But if you do change the names of the calculators, pls still add in a M3GNetCalculator(MGLCalculator) for backwards compatibility for now. We can remove it at a later date.

lbluque:
I updated the code to have a single `Potential` class and simply check the return value from it's model.

kenko911:
@lbluque Sorry for the late reply. The code looks fine to me. However, I am not sure if the def forward in Potential class should output other atomwise properties like charges and spins. What do you think @shyuep and @lbluque since now the . Would it possible to create another function in the Potential class to output atomwise properties instead of breaking the energy, force, stress, hessian calculations into two parts?

shyuep:
I think this is fine. The forward method in a Module is supposed to output predictions. Things like forces and stresses are in fact site wise predictions. Magmom is just another site-wise prediction. I will merge and do a bit of cleanup.

PR #133, opened 2023-08-14
use register buffer in Potential in LightningPotential and PES

lbluque:
My understanding from their docs is that pytorch lightning recommends using `register_buffer` for tensors that are model attributes that are not optimized, and that calls of `to(device)` should be avoided since the lightning module "knows what device it is on".

I have checked that using register_buffer works for training on GPU successfully, but I am not an expert on this so please have a look.

https://lightning.ai/docs/pytorch/stable/accelerators/accelerator_prepare.html

### DIFF
--- a/matgl/apps/pes.py
+++ b/matgl/apps/pes.py
@@ -56,8 +56,10 @@ class Potential(nn.Module, IOMixIn):
         else:
             self.element_refs = None
 
-        self.data_mean = torch.tensor(data_mean) if data_mean is not None else torch.zeros(1)
-        self.data_std = torch.tensor(data_std) if data_std is not None else torch.ones(1)
+        data_mean = torch.tensor(data_mean) if data_mean is not None else torch.zeros(1)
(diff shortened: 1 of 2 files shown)


kenko911:
Hi @lbluque, thank you very much for your information, It would be helpful if register_buffer can effectively avoid using .todevice when using multi-gpus training. I will test with your version and let you know if it also works for multi-gpus training

shyuep:
I agree. I already said before - use `with torch.device("cuda")` or `torch.set_default_device` in the end point code. MatGL should contain ZERO `to(device)` calls.

shyuep:
I had to switch off register buffer for data_mean and data_std. These are actual variables needed that are serialized in model.

shyuep:
Also, I changed the default stress_weight and site_wise_weight to 0.0. A 0 weight basically means ignore this feature? I don't think there is a need to specify None for a weight.

PR #147, opened 2023-08-24
Keep the data type consistent in dgl graphs to be float32 and int32.

kenko911:


shyuep:
This is way too much overengineering. There doesn't need to be a class called DataType. If you look at how the numpy or torch does it, it is just a module.  E.g., np.float32. They don't do np.DataType.float32.

shyuep:
I implemented a solution and pushed it. I suggest you look at it

PR #153, opened 2023-08-29
set pbc_offsift and pos as float64

lbluque:
I set `edata["pbc_offshift"]` and `ndata["pos"]` as float64 to correctly handle creating line graphs when some bond lengths are right at the cutoff.

I added a test for this edge case with a made-up structure.


shyuep:
Thanks.

PR #157, opened 2023-09-01
Added the united test for hessian in test_ase.py to improve the coverage score

kenko911:


kenko911:
coverage score increases :).

PR #158, opened 2023-09-01
AtomRef Updates

lbluque:
A few edits to the `AtomRef` class:

* added `max_z` as optional kwarg for cases where offsets have not been fitted yet.
* Removed all use of `numpy`
* Use `lstsq` instead of explicitly pseudo-inverse
* Made `property_offset` a buffer to avoid errors when running on GPU and with lightning modules
* Save precomputed identity matrix for onehot vectors as buffer.

### DIFF
--- a/matgl/layers/_atom_ref.py
+++ b/matgl/layers/_atom_ref.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import dgl
-import numpy as np
 import torch
 from torch import nn
 import matgl
--- a/matgl/layers/_atom_ref.py
+++ b/matgl/layers/_atom_ref.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import numpy as np
 import dgl
 import torch
 from torch import nn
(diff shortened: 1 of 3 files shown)


lbluque:
The errors have to do with "Bad serialized model". I don't really understand how to fix those.

I also keep getting this error when running pre-commit:
```
error: TOML parse error at line 69, column 3
   |
69 |   "PERF203", # try-except-in-loop
   |   ^^^^^^^^^
Unknown rule selector: `PERF203`
```
Any ideas on how to get rid of it?

shyuep:
I think it is a version of ruff that is the problem for the PERF. For bad serialized model, I think your fork may have cached an old model. Maybe mgl clear --yes command to remove all models first?

lbluque:
Thanks @shyuep ! 

The "bad serialized" error was actually from my changes. I had to do a few things to change it:
- use `strict=False` when using `load_state_dict` in `IOMixin.load`.
- check if `property_offset` is a list or ndarray in `AtomRef` constructure for backcompat with pretrained models.


I also added a raise from the `BaseException` leading to the "Bad serialized" error because otherwise it can be unclear what the error actually is.

Pls have a look and lmk your thoughts

shyuep:
Thanks. I think this is fine. I will merge.

PR #168, opened 2023-09-19
Avoiding crashes for PES training without stresses and update pretrained models

kenko911:
## Summary

Major changes:

Adding an option for PES training without stresses. Moreover, registering some variables in pes.py as buffer to avoid crashes when using multiple GPUs training and update the pretrained M3GNet-PES accordingly

## Checklist

- [ X] Google format doc strings added. Check with `ruff`.
- [ X] Type annotations included. Check with `mypy`.
- [ X] Tests added for new features/fixes.
- [ X] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
@kenko911 The tests are failing.

kenko911:
Hi @shyuep, I managed to fix all the united tests except the test_integration.py. I changed the variable data_mean from None to 0.0 in  model.pt from the pretrained M3GNet-MP-2021.2.8-PES model so that it prevents crashes for the new changes. However, the test_integration.py checks the available pretrained models based on loading the older model version from /home/runner/.cache/matgl/M3GNet-MP-2021.2.8-PES/model.pt and therefore causes the error.

shyuep:
No. The code clears the cache before any tests are run. I also already pushed a fix that allows None for data_mean previously. Your code fails because you didn't code in a way that ensures backward compatibility. There IS a simple code fix. Read my code to understand what to do and then push a fix.

PR #170, opened 2023-09-21
Mlp

lbluque:
## Summary

Major changes:

Edited the MLP code to make it easier to introduce new layers (such as dropout, norm, etc).

## Checklist

- [X] Tests added for new features/fixes.
- [X] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

```sh
pip install -U pre-commit
pre-commit install
```


lbluque:
Unfortunately, this breaks loading the pre-trained models because the ModuleList in the MLP class no longer includes the activation, so parameters in the state dict can not be loaded.

If it is too much unnecessary work to fix loading pre-trained models, I can write new MLP/gMLP classes, but this will be duplicated.

shyuep:
I am a bit confused. I think adding the flexibility for dropout etc is good. But what is wrong with adding these to the ModuleList?

lbluque:
The main motivation is to allow different additional inputs in forward. In particular I would like to be able to pass the graph to some layers. This can be done using the module list, but that will probably need something like an `isistance` or other introspection check in the forward method.

lbluque:
However, I think more importantly the fact that I am re-using the `MLP` class to avoid duplication in the `GatedMLP` breaks the pre-trained models as well.

shyuep:
I understand that. But if you are writing it as a module list, like what was in the original MLP class, the state dict would still be the same for those that do not use those features. I am trying to understand if there is a significant advantage to doing it this way, which causes all previous models to break and we have to rebuild the state dict. It is not trivial to redo the state dict.

lbluque:
> I understand that. But if you are writing it as a module list, like what was in the original MLP class, the state dict would still be the same for those that do not use those features. I am trying to understand if there is a significant advantage to doing it this way, which causes all previous models to break and we have to rebuild the state dict. It is not trivial to redo the state dict.

I see.

In `MLP` I can add the functionality by adding to the `ModuleList` and editing the forward method accordingly.

However I tried changing the `Sequential` in `GatedMLP` to a `ModuleList` to do the same, but that also breaks pre-trained models. 

The only solution I can think of to avoid breaking pre-trained models is to have a separate class.

lbluque:
I'll just go ahead and write different classes if that's ok.  I'll close this PR and do it in a new one.

shyuep:
@lbluque I prefer not to have code duplication. Let's keep this. For MLP, let's write it in a backward compatible manner. I am unsure why it would not be possible to do the same with the gatedMLP?

lbluque:
The problem with `GatedMLP` is that it is implemented with `nn.Sequential` and I am getting errors when swapping it to `nn.ModuleList`. 

Even if I try to evaluate each layer in `nn.Sequential` explicitly in a loop (which would be improper use of the class anyway), I also get errors in the pretrained models.

shyuep:
@kenko911 Pls work with @lbluque to sort this out. If we need to rebuild the statedict, this has to be done in a way that doesn't require retraining all the models.

PR #178, opened 2023-10-02
Directed line graph

lbluque:
## Summary

Implements constructing a directed line graph, where edges in the original graph have an edge in the line graph if the dst node of the first edge matches the src node of the second. i.e. the following directed edges x = (u,v) and y = (v,w) will have a directed edge in the line graph (x, y).

Implemented:
- `create_directed_line_graph`
- `ensure_directed_line_graph_compatibility` to be used when a graph and line graph are passed (ie when training)
- `prune_edges_by_features` to prune graphs according to some feature.

I had to adapt `compute_theta` to handle the "directed" case, bc in this case one of the bond vectors must be flipped to get the correct angle and not its complement.

## Checklist

- [X] Google format doc strings added. Check with `ruff`.
- [X] Type annotations included. Check with `mypy`.
- [X] Tests added for new features/fixes.
- [X] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
I would actually do it by simply add a kwarg called `is_directed` to `create_line_graph` which defaults to False. It should be simple to do based on your current implementation.

lbluque:
@shyuep made the changes you suggested, pls have a look when you can

shyuep:
Thanks

PR #184, opened 2023-10-31
log the # of parameters when loading the model

JiQi535:
When loading a model, I think it is helpful to log the # of parameters and the path of the stored model.


shyuep:
This is the wrong way to use logging. A logger level shouldn't be set in the code. It should be set in your *script* when you actually use it. You may want to have the # of parameters printed, but other people may not care to see it.

Issue #193, opened 2023-11-17
[Bug]: Urgent! ValueError: Bad serialized model or bad model name


janosh:
### Version

0.8.5

### Which OS(es) are you using?

- [ ] MacOS
- [ ] Windows
- [ ] Linux

### What happened?

`matgl` is holding up the whole MP stack with this error

> ValueError: Bad serialized model or bad model name. It is possible that you have an older model cached. Please clear your cache by running `python -c "import matgl; matgl.clear_cache()"`

```
/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/matgl/utils/io.py:213:
```

It's affecting `pymatgen`, `atomate2`, `matcalc`, `emmet`, ...
Please add tests for whatever broke so this doesn't happen again!

- https://github.com/materialsproject/pymatgen/pull/3477
- https://github.com/materialsproject/atomate2/pull/614
- https://github.com/materialsproject/atomate2/pull/615
- https://github.com/materialsvirtuallab/matcalc/pull/11

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shyuep:
Erm, I think the error message says it all? @kenko911 You need to change the model number when you make internal changes like the stress computation. Otherwise the loading of the models can be wrong.

janosh:
How come there's no test that caught this?

ucarh:
I had this error with python 3.12, downgrading it to 3.9 solved the issue.

shyuep:
DGL (the underlying dependency) is not even available on py3.12. In general, we would recommend using Py3.10 (or at most one version below the latest Python, which usually has bugs as well as a lot of unsupported dependencies).

PR #194, opened 2023-11-17
model version for Potential class is modified

kenko911:
## Summary
The model version of the Potential class is changed from 1 to 2. This prevents the wrong loading of the model.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/dev/refactor.py
+++ b/dev/refactor.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 
 import torch
-
 from matgl.models import MEGNet
 
 # model_path = "pretrained_models/MEGNet-MP-2019.4.1-BandGap-mfi"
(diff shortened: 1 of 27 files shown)


janosh:
@kenko911 What do you need from me?

kenko911:
Hi @janosh, thanks for reporting the error. I just changed the model version and I would like to know if it fixes the error you reported in  #193?

janosh:
@shyuep already fixed the error in [v0.9.0](https://github.com/materialsvirtuallab/matgl/releases/tag/v0.9.0).

shyuep:
@janosh Yes it is fixed. But this is more for future-proofness. We want a proper versioning of model so that the cache invalidation works properly. It is not related to the tests in pymatgen/downstream codes.

PR #212, opened 2024-01-13
Add output_layer argument to M3GNet forward method to return intermediate layer outputs

JiQi535:
## Summary
In this PR, the `forward` method of M3GNet is modified so that intermediate outputs in "embedding", "gc_1", "gc_2", "gc_3", and "readout" layers can be extracted, while the "final" layer output stays as the default output. This allows the usage of M3GNet intermediate outputs as atom, bond, and structure features. Documentations and tests are provided.

Major changes:

1. Added the argument of `output_layer` into the `forward` method.
2. Added comprehensive tests for the `predict_structure` method.
3. Added the argument of `output_layer` into the `predict_structure` method.
4. Edited the variable names in the readout layer of M3GNet to be more precise. For example, the `node_vec` is renamed to `field_vec`, as this variable can either be node feature or edge feature instead of just node feature, depending on the argument `field`.

## Todos

Follow-up analysis for the effectiveness of atom features.

## Checklist

### DIFF
--- a/src/matgl/models/_m3gnet.py
+++ b/src/matgl/models/_m3gnet.py
@@ -256,7 +256,7 @@ class M3GNet(nn.Module, IOMixIn):
                 edge_feat,
             )
             edge_feat, node_feat, state_feat = self.graph_layers[i](g, edge_feat, node_feat, state_feat)
-            if i + 1 in output_layer:
+            if f"{i + 1}" in output_layer:
                 return node_feat, edge_feat, state_feat
         g.ndata["node_feat"] = node_feat
         g.edata["edge_feat"] = edge_feat
@@ -302,4 +302,6 @@ class M3GNet(nn.Module, IOMixIn):
         g.ndata["pos"] = g.ndata["frac_coords"] @ lat[0]
         if state_feats is None:
             state_feats = torch.tensor(state_feats_default)
-        return self(g=g, state_attr=state_feats, output_layer=output_layer).detach()
+        if output_layer == "final":
+            return self(g=g, state_attr=state_feats).detach()
+        return self(g=g, state_attr=state_feats, output_layer=output_layer)
--- a/src/matgl/models/_m3gnet.py
+++ b/src/matgl/models/_m3gnet.py
@@ -179,7 +179,10 @@ class M3GNet(nn.Module, IOMixIn):
             input_feats = dim_node_embedding if field == "node_feat" else dim_edge_embedding
             if readout_type == "set2set":
                 self.readout = Set2SetReadOut(
-                    in_feats=input_feats, n_iters=niters_set2set, n_layers=nlayers_set2set, field=field
+                    in_feats=input_feats,
+                    n_iters=niters_set2set,
+                    n_layers=nlayers_set2set,
+                    field=field,
                 )
                 readout_feats = 2 * input_feats + dim_state_feats if include_state else 2 * input_feats  # type: ignore
             else:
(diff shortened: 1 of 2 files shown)


shyuep:
I think it is an extremely bad practice to put this in the predict_structure method. The predict_structure should return the expected target of the model. You should put it in a different method.

kenko911:
Hi @JiQi535, thanks for your PR. I agree that we should have another class to extract features from structures instead of just putting in the function predict_structure, which is mainly for property prediction. Maybe we can call the new function "extract_features" and then we can also remove the unnecessary "final" option in the output_layer?

JiQi535:
Thanks for the comments @shyuep @kenko911 . I have added a different method, `featurize_structure`, and the `predict_structure` method has been changed back to its original format to only return the expected target of the model. 

By default,  `featurize_structure` now returns the "readout" layer output and the "gc_3" layer output of intensive and extensive M3GNet models, respectively. Also, tests are updated, and a ValueError is raised if `output_layer` is not one of the allowed names in ["embedding", "gc_1", "gc_2", "gc_3", "readout", "final"].

shyuep:
There is just way too much code and this is poor software design.
1. featurize_structure should simply return a dict of all intermediate and final layers. This way, there is no need to recompute every time you need one layer.
2. predict_structure should call featurize_structure in some way. Even better, the forward method actually stores the computed layers and you can access it immediately.

JiQi535:
@shyuep The two suggestions have been realized in the updated code. My edits do not change the default output of `forward` method, so that other related codes, e.g., the `Potential` class, do not need change. This requires the addition of an argument in the forward method, i.e., `return_all_layer_output`. This implementation also does not increase the memory consumption of the `forward` method when `return_all_layer_output` is False as default.

src/matgl/models/_m3gnet.py
```diff
@@ -251,35 +259,60 @@ def forward(
...
             graph_converter: Object that implements a get_graph_from_structure.
+            output_layers: List of names for the layer of GNN as output. Choose from "bond_expansion", "embedding",
+                "three_body_basis", "gc_1", "gc_2", "gc_3", "readout", and "final". By default, all M3GNet layer
+                outputs are returned.
 
         Returns:
-            output (torch.tensor): output property
+            output (dict): M3GNet intermediate and final layer outputs for a structure.
         """
+        allowed_output_layers = [
```
shyuep, line 301: This cannot be hard coded. So if you have more than three GCLs or less than 3 GCLs, this code fails. Think about what kind of code you are creating. Are we saying M3GNet ALWAYS have 3 GCLs?

src/matgl/models/_m3gnet.py
```diff
@@ -215,16 +215,19 @@ def forward(
         g: dgl.DGLGraph,
         state_attr: torch.Tensor | None = None,
         l_g: dgl.DGLGraph | None = None,
+        return_all_layer_output: bool = False,
```
shyuep, line 218: Why is this needed? There is no compute cost to returning all layers right? Why not return all and then use the relevant layer?

src/matgl/models/_m3gnet.py
JiQi535: I will address this in next commit.

src/matgl/models/_m3gnet.py
JiQi535: If we return all layer outputs, the default behavior of forward method will be changed.

PR #225, opened 2024-02-10
Ensure `free_energy` property is `float`

Andrew-S-Rosen:
This is a follow-up to #203.

The `Calculator.results` dictionary from an M3GNet run yields something like `'free_energy': array(-4.0938973, dtype=float32)` for the free energy property. This causes some havoc in places (e.g. monty jsanitization) since the array isn't iterable. It doesn't have a defined `len` without the brackets. We should really make this a `float` instead anyway. That's what I do in this PR.

### DIFF
--- a/src/matgl/ext/ase.py
+++ b/src/matgl/ext/ase.py
@@ -177,7 +177,7 @@ class M3GNetCalculator(Calculator):
             energies, forces, stresses, hessians = self.potential(graph, lattice, state_attr_default)
         self.results.update(
             energy=energies.detach().cpu().numpy().item(),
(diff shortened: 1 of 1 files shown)


kenko911:
Hi @Andrew-S-Rosen, thanks for your PR! I noticed that the current AdvancedSoft MatGL-LAMMPS interface breaks due to the changes for the datatype of potential_energy() from numpy.ndarray into float (see line 105 in https://github.com/advancesoftcorp/lammps/blob/based-on-lammps_2Aug2023/potentials/M3GNET/matgl_driver.py). I will do a PR to AdvancedSoft to fix the bug.

PR #227, opened 2024-02-11
Merge the changes from the branch tko

kenko911:
## Summary
Merge the major changes from branch tko including new model architectures, refactoring variables and unifying training frameworks.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


src/matgl/ext/ase.py
```diff
@@ -119,8 +119,8 @@ def get_graph(self, atoms: Atoms) -> tuple[dgl.DGLGraph, torch.Tensor, list]:
         return g, lat, state_attr
 
 
-class M3GNetCalculator(Calculator):
```
shyuep, line 122: Why cause unnecessary backward incompatible changes? You can add a line `M3GNetCalculator = PESCalculator` to make sure that existing codes using these work. Right now, you are going to cause all kinds of havoc in pymatgen, matcalc, etc.

src/matgl/layers/_graph_convolution.py
```diff
@@ -243,12 +246,12 @@ def __init__(
         state_update_func: Module | None,
     ):
         """Parameters:
-        include_state (bool): Whether including state
```
shyuep, line 246: Again, you cannot change variable names on a whim. This can be done in a backward compatible manner. But more importantly, did you even update the version numbers?

src/matgl/ext/ase.py
kenko911: @shyuep, I just added the M3GNetClass, which inherits PESCalculator for backward compatibility.

PR #259, opened 2024-05-11
Ensure `free_energy` property is a `float` and not `np.array(float)`

Andrew-S-Rosen:
The `PESCalculator.results["free_energy"]` property should be formatted as a `float` but is currently formatted as a `np.array(float)`. For instance, `array(-4.0938973, dtype=float32)`. Since this is not the expected ASE type and because it is not iterable, this can cause a variety of unexpected problems. https://github.com/materialsvirtuallab/monty/issues/668 is just one example.

I have ensured the `free_energy` property is returned as a `float` in this PR.


kenko911:
Hi @Andrew-S-Rosen, thanks for your changes! Merged.

PR #403, opened 2024-10-22
Allow missing label in MGLDataset

BowenD-UCB:
## Summary
This PR allows loss function calculation in MGLDataset.
Now we're able to train with the dataset of partial labels.
The default behavior of the training and loss function is `allow_missing_labels: bool = False`
So this PR does not influence the performance of previous working codes.

Major changes:

- `src/matgl/utils/training.py`
- `tests/utils/test_training.py`

Example:
```
energies[2] = np.nan
forces[4] = (np.nan * np.ones((len(structures[4]), 3))).tolist()
stresses[6] = (np.nan * np.ones((3, 3))).tolist()
magmoms[8] = (np.nan * np.ones((len(structures[8]), 1))).tolist()

dataset = MGLDataset(
    structures=structures,
    converter=converter,
    labels={
        "energies": energies,
        "forces": forces,
        "stresses": stresses,
        "magmoms": magmoms,
    },
)
```
The above dataset with some missing labels should be now trainable

### DIFF
--- a/src/matgl/utils/training.py
+++ b/src/matgl/utils/training.py
@@ -462,6 +462,12 @@ class PotentialLightningModule(MatglLightningModuleMixin, pl.LightningModule):
 
         """
         # labels and preds are (energy, force, stress, (optional) site_wise)
+        if self.allow_missing_labels:
+            for i, label in enumerate(labels):
+                valid_value_indices = ~torch.isnan(label)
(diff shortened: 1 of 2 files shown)


kenko911:
Hi @BowenD-UCB, thanks for the PR and it looks great!

Issue #572, opened 2025-04-01
[Bug]: README has inaccurate torch version


Andrew-S-Rosen:
### Email (Optional)

CC'ing @naisargi-goyal

### Version

1.2.1

### Which OS(es) are you using?

- [ ] MacOS
- [ ] Windows
- [x] Linux

### What happened?

In the README, it says:

```
pip install torch==2.2.1 torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118
pip install dgl -f https://data.dgl.ai/wheels/cu118/repo.html
pip install dglgo -f https://data.dgl.ai/wheels-test/repo.html
```

However, the `pyproject.toml` of matgl lists `torch<=2.2.0` as a dependency. Therefore, there is a conflict if the user installs PyTorch as instructed in the README.

### Code snippet

```python

```

### Log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #574, opened 2025-04-01
Update README.md to have clearer torch installation instructions

Andrew-S-Rosen:
## Summary

Closes https://github.com/materialsvirtuallab/matgl/issues/572. The installation instructions now recommend `torch==2.2.0` instead of `torch==2.2.1` since the latter is not supported by matgl's `pyproject.toml`. I also removed `torchvision` and `torchaudio`, which are not needed here, and I switched to `cu121` in place of `cu118` since CUDA 12.1 is the highest version supported by the current dependencies.

## Summary by CodeRabbit

- **Documentation**
  - Updated the installation instructions to reflect revised recommended versions for PyTorch and DGL, ensuring enhanced compatibility with newer CUDA requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ and DGL. The basic instructions are given below, but it is recommended that you 
 run into any problems.
 
 ```shell
-pip install torch==2.2.1 torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118
-pip install dgl -f https://data.dgl.ai/wheels/cu118/repo.html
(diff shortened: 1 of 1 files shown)


shyuep:
Many thanks!

shyuep:
Also, just a note that we have release docker images with and without GPU as well as the LAMMPS interface.

Andrew-S-Rosen:
Ah, wonderful! That will help a fair bit. We were trying to get ourselves out of dependency hell for a little while...

PR #580, opened 2025-04-10
Add upper python version specifier

DanielYang59:
## Summary

- Add upper python version specifier

Using an explicit upper-bound specifier helps prevent package managers (e.g., pip) from installing the package on unsupported future Python versions 

## Summary by CodeRabbit

- **Chores**
	- Updated project compatibility to support Python versions 3.10 through less than 3.13.
	- Explicitly added support for Python 3.12.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ authors = [
 ]
 description = "MatGL is a framework for graph deep learning for materials science."
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.13"
 keywords = [
     "materials",
     "interatomic potential",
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks.

DanielYang59:
No problem :D

DanielYang59:
@shyuep I realize it might be a bad idea to set a upper bound in general, see [this very detailed discussion on this](https://discuss.python.org/t/requires-python-upper-limits/12663) pointed out by Ahmed Ilyas.

So the TL;DR for this is: Python seldomly introduce backwards incompatible change/deprecate any API, so without an upper bound, the package manager would try to install on an untested Python version, which would either just work (most likely) or error out. But with an upper bound, we're basically just give up the chance to give it a try.

I would open a PR to drop this upper bound

PR #581, opened 2025-04-16
Remove duplicate dependencies in `pyproject.toml`

Andrew-S-Rosen:
There were duplicated dependencies in the `pyproject.toml`. I did some clean up.


shyuep:
These were not duplicates. They are platform specific. I was trying to allow Ubuntu users to use later versions of DGL, torch and torch data rather than being tied to the Mac/Win versions.

PR #583, opened 2025-04-20
Allow disable BondGraph in CHGNet

bowen-bd:
## Summary

By setting `threebody_cutoff = 0`, CHGNet should skip all bond graph initialization and forward passes. 
This allows for more flexibility, especially for speed testing

The changes do not add additional initialization keys and are backward compatible with previous checkpoints


shyuep:
Thanks,

PR #591, opened 2025-05-04
`_create_directed_line_graph` error handling

bowen-bd:
## Bug Report

Recently, we found CHGNet inference can trigger very low-level line graph creation bug on certain structure during MD.
```python
pot = matgl.load_model('CHGNet-MatPES-r2SCAN-2025.2.10-2.7M-PES')
calculator = PESCalculator(potential=pot)
structure = Structure.from_file(filename='./bug_structure_4.30_POSCAR')
poscar_atoms = structure.to_ase_atoms()
poscar_atoms.calc = calculator

print(poscar_atoms.get_potential_energy())
```
A simple forward will trigger a RuntimeError and stop the whole simulation like a Molecular Dynamics, see below
```text
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
File ~/Packages/matgl/src/matgl/models/_chgnet.py:353, in CHGNet.forward(self, g, state_attr, l_g)
    351 if self.use_bond_graph:
    352     if l_g is None:
--> 353         bond_graph = create_line_graph(g, self.three_body_cutoff, directed=True)
    354     else:
    355         # need to ensure the line graph matches the graph
    356         bond_graph = ensure_line_graph_compatibility(g, l_g, self.three_body_cutoff, directed=True)

File ~/Packages/matgl/src/matgl/graph/compute.py:92, in create_line_graph(g, threebody_cutoff, directed)
     90 graph_with_three_body = prune_edges_by_features(g, feat_name="bond_dist", condition=lambda x: x > threebody_cutoff)
     91 if directed:
---> 92     lg = _create_directed_line_graph(graph_with_three_body, threebody_cutoff)
     93 else:
     94     lg = _compute_3body(graph_with_three_body)

File ~/Packages/matgl/src/matgl/graph/compute.py:263, in _create_directed_line_graph(graph, threebody_cutoff)
    261 lg_src_ns = incoming[not_self_edge].nonzero()[:, 1].squeeze()
    262 lg_dst_ns = edge_inds_ns.repeat_interleave(num_edges_per_bond[not_self_edge])
--> 263 lg_src[n:], lg_dst[n:] = lg_src_ns, lg_dst_ns
    264 lg = dgl.graph((lg_src, lg_dst))
    266 for key in pg.edata:

RuntimeError: The expanded size of the tensor (1248) must match the existing size (1249) at non-singleton dimension 0.  Target sizes: [1248].  Tensor sizes: [1249]
---------------------------------------------------------------------------
```

## Observation with the BUG
This is a very concerning and low-level bug, we suspect it's coming from certain numerical issues of the bonds in certain atomic arrangements. But here's my current observations:
1. This bug occurs very, very rarely. When we trained CHGNet on MPtrj and MatPES, it never occurred for those millions of structures.
2. This bug is very sensitive to numerical precision. When I perturb the structure for a tiny bit, it's gone. From my testing, a `cif` to `POSCAR` conversion of the same structure (which in theory involves some float precision changes), can make the bug go away.
3. I spent a few days digging into what could fundamentally go wrong with the logic of the line graph creation and could not figure out a clean solution. The bug is not reproduceable unless providing the structure with the exact float digits. I included the erroneous structure in the pytest.

## Solution
 So, here is a dirty solution to this rare problem for now:
1. I added an error handler when the line graph construction triggers RuntimeError, this is by adding a tiny noise `1e-6` to line graph converter.
2. The error handler is only added to directed line graph converter, so only CHGNet is affcted.
3. By default, the error handler is always on for CHGNet now. But this should really affect only one in a million cases of inference.
4. I added some pytests for both 
a. reproducing the line graph bug
b. showing that the error handler only adds 1e-7 amount of inconsistency in force prediction when it triggers, which should not affect any simulation accuracy

Comments are welcome if any others have observed and suffered from similar bug before, thanks!


bowen-bd:
This is probably the most annoying bug I encountered this year :sweat_smile:
I wish this solution can deal with most of the cases and make the use case more robust.

bowen-bd:
@shyuep Hi Shyueping! Can you please merge this when you have time? thanks!

shyuep:
Done. Thanks!

PR #633, opened 2025-08-12
Pretrained molecular potentials from MatGL paper and additional thermostats for ASE MD added

kenko911:
## Summary
The molecular potentials pretrained on the subset of ANI-1x dataset is included. Moreover, more stable thermostats (Nose hoover chain) for both NVT and NPT are also added.

## Todos

If this is work in progress, what else needs to be done?

- feature 2: ...
- fix 2:

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


Andrew-S-Rosen:
Sidenote: Potential breaking change [here](https://github.com/materialsvirtuallab/matcalc/actions/runs/16916505594/job/47931624345?pr=93) that I think may be related to the matgl release.

kenko911:
Hi @Andrew-S-Rosen, thank you very much for your comments! It seems that the default values of taut and taup are similar to those of tdamp and pdamp. Perhaps it would be simpler to remove pdamp and tdamp and just use taut and taup for all ensembles. Regarding MatCalc, it seems that the unit tests do not work for molecular potentials in general.

PR #694, opened 2025-11-30
[WIP] Enable inference on GPU for pyg ASE calculator

mjwen:
## Summary

Major changes:

- Allows inference to run on GPU for pyg ASE calculator. This can be done via the newly added `device` argument to the `PESCalculator`. 

## Todos

- Just did a quick fix to make it work for `PESCalculator`. Should update other calculators. 

## Questions 

- Potentials like `TensorNet` expect `state_attr` to be a torch.Tensor. But `Atoms2Graph` returns a `list | np.ndarray`. Currently, I've changed `Atoms2Graph` to return a `torch.Tensor`, but this might not be a good idea since `Atoms2Graph` can be used in other places. Alternatively, the type conversion can be done within the `PESCalculator.calculate` method. Any suggestions? 

## Checklist

- [ ] Google format doc strings added. Check with `ruff`.
- [ ] Type annotations included. Check with `mypy`.

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/src/matgl/ext/_ase_pyg.py
+++ b/src/matgl/ext/_ase_pyg.py
@@ -14,6 +14,7 @@ import ase.optimize as opt
 import numpy as np
 import pandas as pd
 import scipy.sparse as sp
+import torch
 from ase import Atoms, units
 from ase.calculators.calculator import Calculator, all_changes
 from ase.filters import Filter, FrechetCellFilter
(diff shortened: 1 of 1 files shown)


elemynt-ong:
Thanks. But as far as I know, there is no need to pass the device to the Potential class. E.g.,

```python
if torch.cuda.is_available():
    with torch.device('cuda'):
        <do something>
```

will automatically set the default device to CUDA everywhere in the code and allow you to run on GPUs.

mjwen:
Thanks!

My bad. I see it in the tutorial docs now. I should have read it more carefully :-)

Issue #702, opened 2025-12-31
Further clarification of stress units in `PESCalculator`


Andrew-S-Rosen:
I think the handling of stress units in `PESCalculator` must be made much clearer.

Currently, the `Relaxer` class in matgl [handles the conversion internally](https://github.com/materialyzeai/matgl/blob/7100676ebdc81f2f0a3da7d431af167d63be6746/src/matgl/ext/_ase_pyg.py#L253) by setting `stress_weight = 1 / 160.21766208`. The `MolecularDynamics` class in matgl [handles the conversion internally](https://github.com/materialyzeai/matgl/blob/7100676ebdc81f2f0a3da7d431af167d63be6746/src/matgl/ext/_ase_pyg.py#L454-L456) by setting `stress_weight = 1.0` but forcing `stress_unit="eV/A3"`. And the `load_matgl` [method in matcalc](https://github.com/materialyzeai/matcalc/blob/e77af3abeec31fce74f96a661778a1579f264eec/src/matcalc/utils.py#L153-L175) sets `stress_unit="eV/A3"` for the user.

However, if someone instantiates a `PESCalculator` directly (e.g. after a `matgl.load_model()` call) and uses pure ASE for everything else (or even a matcalc utility), there is a very good chance they will completely miss anything related to `stress_weight` or `stress_unit`. I agree with the original poster that it is rather unintuitive and, while I don't think it should be changed in order to maintain backward compatibility, I think there needs to be substantially clearer messaging for the end user.

Currently, there is no direct statement in the README or the docstring of `PESCalculator` that explicitly warns the user something like:
> If you directly instantiate the `PESCalculator` and plan to use it with ASE, you must set `stress_weight=ase.units._GPa_to_eV_per_A3` or `stress_unit="eV/A3"` to ensure the units of stress are in eV/A^3 instead of GPa.

The `PESCalculator` documentation does say the default unit of stress is GPa, but this is counter-intuitive because it is an ASE calculator yet uses ASE-incompatible units out of the box. Also, most users are probably not aware of the implications of this.

My suggestion would be:
1. Make the documentation even clearer in the `PESCalculator` docstring as described above.
2. Ideally, one would handle the unit conversion in the same way between `Relaxer` and `MolecularDynamics` to prevent confusion, but this would probably be a breaking change to default values.  
[...]

PR #703, opened 2026-01-01
Improve the handling of stress units in PESCalculator

kenko911:
## Summary
To address concerns regarding the misuse of stress_units and stress_weight in PESCalculator (see #702), the following changes have been implemented.
Major changes:
- Fix 1: Clarify and expand the PESCalculator docstring to explicitly document supported stress units (GPa and eV/Å³), with "GPa" specified as the default.

- Fix 2: Add explicit Warning, Note, and Error messages in PESCalculator to ensure users are informed of the currently configured stress units and to prevent ambiguous or inconsistent unit usage.

- Fix 3: Deprecate stress_weight in Relaxer and MolecularDynamics, as stress units must be set to "eV/Å³" for physically consistent behavior. PESCalculator retains support for "GPa" solely for backward compatibility with the AdvanceSoft LAMMPS interface and other packages.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/src/matgl/ext/_ase_dgl.py
+++ b/src/matgl/ext/_ase_dgl.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import collections
 import contextlib
 import io
+import logging
 import math
 import pickle
 import sys
@@ -44,6 +45,8 @@ if TYPE_CHECKING:
     from ase.optimize.optimize import Optimizer
 
     from matgl.apps._pes_dgl import Potential
+
+logger = logging.getLogger(__name__)
 
 
 class OPTIMIZERS(Enum):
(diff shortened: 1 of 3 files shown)


src/matgl/ext/_ase_dgl.py
```diff
@@ -185,7 +187,23 @@ def __init__(
...
+        self.conversion_factor = conversion_factor * stress_weight
+        if self.conversion_factor == 1.0:
+            print(
+                "Warning: The stress unit is now in GPa. Please set the stress_unit to be "
+                "'eV/A3' if you want to use PESCalculator for other ASE applications."
+            )
+        elif math.isclose(
+            self.conversion_factor, units.GPa / (units.eV / units.Angstrom**3), rel_tol=1e-4, abs_tol=1e-6
+        ):
+            print("Note: The stress unit is now in eV/A^3, which is the correct unit for ASE Calculator.")
```
Andrew-S-Rosen: It may make more sense to use the Python `logging` utility so the end user can control the log level. I imagine the user probably won't want to see `Note: The stress unit is now in eV/A^3` every time they run a job (once they are confident things are working as expected).

If this is done, then the first message should probably be `WARNING` level and the latter `INFO` level.

Andrew-S-Rosen (commented):
@kenko911, thank you very much for putting this together. I hope that these changes, or ones similar to it, will help prevent errors in the future. Unfortunately, our group was significantly impacted by this and so the goal is for others to not fall for the same mistake.

I have one comment below regarding the logging, which you may wish to consider. Otherwise, this seems logical to me.

src/matgl/ext/_ase_dgl.py
kenko911: Thanks for the comments. I have now used logging!

PR #705, opened 2026-01-04
feat: enable returning intermediate crystal features in CHGNet and TensorNet

bowen-bd:
This PR adds support for returning intermediate layer features in CHGNet and TensorNet (both DGL and PyG), similar to the functionality already present in M3GNet.
This helps for consistent behaviour of different models and potentials.

All tests have passed locally across both DGL and PyG backends.


kenko911:
Thanks @bowen-bd!

PR #755, opened 2026-04-18
[WIP] Message aggregation direction issue that causes incorrect energies in layered structures

drakeyu:
## Summary

Major changes: Fixes a bug in the PyG `TensorNetInteraction` layer where messages are aggregated into the wrong set of nodes, causing incorrect energies for asymmetric/layered structures.

This is inconsistent with the DGL implementation, which correctly aggregates into `dst` nodes via `update_all(fn.copy_e, fn.sum)`, and inconsistent with `TensorEmbedding` (line 176) which already correctly uses `edge_index[1]`.

### Fix
Change line 116 in `_graph_convolution_pyg.py` from:
```python
Im, Am, Sm = self.aggregate(messages, edge_index[0], X.size(0))
```
to:
```python
Im, Am, Sm = self.aggregate(messages, edge_index[1], X.size(0))
```
## Todos

retrain the potential

## Checklist

- [ ] Google format doc strings added. Check with `ruff`.
- [ ] Type annotations included. Check with `mypy`.

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ classifiers = [
 ]
 dependencies = [
     "ase",
-    "torch",  
+    "torch",
     "torchdata",
     "lightning<=2.6.1",
     "pydantic",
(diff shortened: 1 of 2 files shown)


shyuep:
You need to properly fix all the tests since those were based on the buggy implementation. Also, add a test to ensure that the bug does not recur.

shyuep:
I implemented the fix in the tnfix branch for now, unless we actually release the retrained models as well as the more permanent fix.

PR #768, opened 2026-05-04
M3gnet pyg

shyuep:
Port of PYG m3gnet and megnet.


esoteric-ephemera:
@shyuep is there no plan to maintain the older M3GNet / DGL-based models? Or port them to PYG?

shyuep:
@esoteric-ephemera What do you mean by older models? You mean the ones that were trained previously? Those contain the bug. That's why we deleted them - it is not good for anyone to use them. That's why it is better to cause failures for those pre-trained models than to let them fail silently. New M3GNet models, which are largely based on the same MatPES dataset, have already been released. See https://huggingface.co/materialyze/M3GNet-PES-MatPES-r2SCAN-2025.2

esoteric-ephemera:
Yes, the 2025.1 MatPES models, and others from previously published papers, like M3GNet trained on MPF or the M3GNet formation energy model used for DIRECT sampling

Since these have been around long enough and have been used, they likely need to be preserved somewhere (HF or GH). They can absolutely come with a warning that they are no longer supported

I also don't see a replacement for the EFORM M3GNet model, only a MEGNet one. Is that what you're recommending for DIRECT sampling now?

shyuep:
Yes, the MPF and formation energy models will be refitted. Probably not the 2025.1 models, since the 2025.2 models are basically the same thing?

esoteric-ephemera:
That's all fine, but shouldn't legacy published models (original M3GNet and its formation energy model) still be available on HF even just for reproducibility?

shyuep:
@esoteric-ephemera They are being refitted. Anyway, I uploaded them to HF so that people can have access to them if needed. I don't really see a huge value in keeping older models that are incorrect.

esoteric-ephemera:
Thanks - it would be good to include the 2025.1 MatPES models as well. Once a method has been published and used to any extent, there should be a persistent copy of it somewhere.

Ex from DFT: [HSE03](https://doi.org/10.1063/1.1564060) was a publication error fixed 3 years later in an [erratum as what we now call HSE06](https://doi.org/10.1063/1.2204597). But since people had already used HSE03 "extensively", most DFT codes preserve both methods

PR #772, opened 2026-05-06
Add per-epoch prediction saving to PotentialLightningModule

YiqingChen524:
New arguments to `PotentialLightningModule` (both DGL and PyG backends):

- `save_predictions` (bool, default `False`): enable saving
- `save_forces` (bool, default `False`): also save per-atom forces
- `save_train_predictions` (bool, default `False`): also save training-set predictions
- `prediction_save_dir` (str, default `"predictions"`): output directory

### DIFF
--- a/src/matgl/utils/__init__.py
+++ b/src/matgl/utils/__init__.py
@@ -1,1 +1,7 @@
 """Various utility methods and classes."""
+
+from __future__ import annotations
+
+from matgl.utils.callbacks import PredictionLogger, add_sample_indices
+
+__all__ = ["PredictionLogger", "add_sample_indices"]
--- a/src/matgl/utils/_training_dgl.py
+++ b/src/matgl/utils/_training_dgl.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, Literal
 import lightning as pl
 import torch
 import torch.nn.functional as F
+from pathlib import Path
 import torchmetrics
 from torch import nn
 
@@ -289,6 +290,10 @@ class PotentialLightningModule(MatglLightningModuleMixin, pl.LightningModule):
         sync_dist: bool = False,
         allow_missing_labels: bool = False,
         magmom_target: Literal["absolute", "symbreak"] | None = "absolute",
+        save_predictions: bool = False,
+        save_forces: bool = False,
+        save_train_predictions: bool = False,
+        prediction_save_dir: str | Path = "predictions",
         **kwargs,
     ):
         """Init PotentialLightningModule with key parameters.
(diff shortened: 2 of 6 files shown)


shyuep (changes requested):
> **Automated PR review generated by Claude (Opus 4.7)** on behalf of @shyuep. Based purely on the diff and CI state — no code was executed locally.

Thanks @YiqingChen524 — useful feature for diagnosing training dynamics. A few blockers and design points before this can merge:

**Blockers**

1. **`lint` is failing** on the head commit (the only red check). Please run `pre-commit run --all-files` locally and push the fixes.

2. **Significant code duplication.** The exact same ~90 lines of hook logic (`on_train_epoch_start/end`, `on_validation_epoch_start/end`, `on_test_epoch_start/end`, plus the buffer-append in `step`) are copy-pasted into both `_training_dgl.py` and `_training_pyg.py`. This is a long-standing problem in this module, but doubling it down across two files makes future maintenance worse. Please extract into either (a) a `PredictionSavingMixin` in a shared module that both `PotentialLightningModule` classes inherit from, or (b) a Lightning `Callback` (the cleaner option — Lightning provides exactly this hook surface, and it lets users opt in via `Trainer(callbacks=[...])` without subclass changes). The `Callback` route also means no breaking-change risk to the constructor signature.

3. **No tests.** Eight new lifecycle methods, four new ctor args, and a non-trivial side-effecting on-disk path, with zero coverage. Please add at least: (i) a smoke test that fits one mini epoch with `save_predictions=True` and asserts the expected `epoch=0000_val.pt` / `train_num_atoms.pt` / `val_indices.pt` files appear with the right shapes; (ii) a test that `save_forces=True` writes `forces_pred`; (iii) a test that the multi-dataloader `ValueError` fires.

**Design issues**

4. **Multi-dataloader validation should fail fast.** Today the `ValueError` is raised in `on_validation_epoch_start` / `on_test_epoch_start`, which crashes the run after potentially expensive setup. Detect in `__init__` (or at least in `on_fit_start`) by inspecting `trainer.val_dataloaders` once, and refuse early.

5. **`forces_pred` buffer is always allocated** even when `save_forces=False`. Minor memory waste and confusing — only populate the key when used.

6. **`hasattr(dataset, "indices")` is a Subset-detection hack.** Use `getattr(dataset, "indices", None) is not None` and document explicitly that this is for `torch.utils.data.Subset` (used by random/manual splits) — otherwise readers will wonder which datasets do/don't have the attribute.

7. **`prediction_save_dir` is normalized to `str` then re-wrapped in `Path` on every hook.** Just store it as `Path(prediction_save_dir)` once in `__init__`. Also: paths under user-configurable directories should be created via `Path.mkdir(parents=True, exist_ok=True)` once in a setup hook, not on every epoch end.

8. **No protection against overwriting prior runs.** If `prediction_save_dir` already contains `epoch=0000_val.pt` from a previous training run, this silently clobbers it. Either add a `version_<N>` subdir, or default `prediction_save_dir` to something under `trainer.log_dir` so each Lightning run is isolated.

9. **Public ctor surface keeps growing.** `PotentialLightningModule.__init__` is already at 20+ kwargs; this adds 4 more. Strong argument for the `Callback` approach in (2) — keeps the module API focused on the loss/optim and pushes I/O concerns out.

10. **Docstring on `on_train_epoch_end`** says "then step the LR scheduler" — but the visible code only calls `super().on_train_epoch_end()`. Either the LR step is in the parent (in which case the comment is misleading because it's parent behavior, not this method's), or it's missing. Please clarify.

Happy to re-review once lint is green, the duplication is factored out, and tests are added.


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #781, opened 2026-05-08
perf: low-risk speedups in graph converters, training loop, and ASE calculator

shyuep:
Five focused, behaviour-preserving speedups identified in the
audit on `main`. No checkpoint format or public-API changes; no
`__version__` bumps required.

## What this changes

| # | File(s) | Change |
|---|---------|--------|
| 2 | `graph/_converters_{pyg,dgl}.py` | Build a `{symbol: idx}` dict once and use `np.fromiter`, replacing the per-atom `element_types.index(...)` call. O(N_atoms * len(element_types)) → O(N_atoms). |
| 3 | `utils/training.py` `loss_fn` | Hoist `valid_labels[0] / valid_num_atoms` and the prediction division out of three metric calls (loss / MAE / RMSE). Same fix in the `symbreak` magmom branch where `-valid_preds[3]` was being recomputed three times. |
| 4 | `utils/training.py` mixin | Remove redundant `torch.set_grad_enabled(True)` from `test_step` and `predict_step`. `PotentialLightningModule.step` already enables grad itself for autograd-based force/stress computation; the outer calls were no-ops for PES models and were preventing `ModelLightningModule` from running under Lightning's default eval-mode optimisations. |
| 6 | `graph/_data_{pyg,dgl}.py` `MGLDataLoader` | Default `num_workers`, `pin_memory`, and `persistent_workers` when the caller didn't pass them, so first-time users aren't silently CPU-bound on GPU training. Explicit kwargs always win. |
| 7 | `ext/_ase_{pyg,dgl}.py` `PESCalculator.calculate` | Detach each output tensor to host once per MD step and reuse the numpy view. Previously `.detach().cpu()` ran twice on the energy tensor (and on stress when `use_voigt=True`), each forcing an extra GPU→CPU sync. |

## Why these and not the rest of the audit

- **Skipped item 1 (vectorised spherical harmonics):** real win, but
  needs careful numerics validation against shipped checkpoints; better as
  a separate PR.
- **Skipped item 5 (CLI lazy imports), 8/9 (sympy/CG cache), 10
  (`scatter_reduce_`), 11 (Hessian vmap), 14 (functional activations):**
  next batch — kept this PR focused on changes that are obviously safe
  and produce no diff in the forward pass output.

## Risk assessment

All five changes are correctness-preserving:

- The converter change is a literal lookup-table rewrite — same `int64`
  values out, same order, same dtype.
- The `loss_fn` hoist computes the same tensor and feeds it into the
  same metric calls; division is associative under the operations
  involved.
- The mixin grad cleanup leaves PES behaviour untouched (`step` still
  enables grad) and only changes non-PES test/predict to honour
  Lightning's eval mode — which is the documented Lightning default.
- The DataLoader defaults are no-ops when the user supplies kwargs and
  match what every PyTorch tutorial recommends.
- The PESCalculator change keeps the numpy outputs bit-identical (same
  detach, same cast, same `full_3x3_to_voigt_6_stress`); it only
  removes redundant transfers.

## Verification

- `uv run ruff check` ✓
- `uv run ruff format --check` ✓
- `uv run mypy -p matgl` ✓ (no issues, 95 source files)
- `uv run pytest` (PYG default backend): **185 passed, 35 skipped**
- `MATGL_BACKEND=DGL .venv_dgl/bin/python -m pytest tests/utils/test_training_dgl.py tests/ext/test_ase_dgl.py tests/ext/test_pymatgen_dgl.py tests/graph/test_data_dgl.py`: **48 passed, 2 skipped** (the 2 skips are pre-existing — CHGNet model needs an update; one MD test is superseded)
- Pre-commit hooks (ruff / format / mypy / codespell) all pass.

## Estimated impact

Order-of-magnitude estimates from the audit; not yet benchmarked
end-to-end.

- Item 2: 5–20% faster graph build on chemically diverse datasets.
- Item 3: 3–8% per training step (depends on metric backend).
- Item 4: 10–20% on validation/test wall time for non-PES `ModelLightningModule`; neutral for PES.
- Item 6: 5–20% throughput on GPU training when users had not
  set `num_workers`/`pin_memory`; neutral when they had.
- Item 7: 5–20% ASE-MD throughput at small/moderate cell sizes where
  Python-side overhead dominates model compute.

A proper `pytest-benchmark` suite (graph build / forward / forward+backward / full step) on a fixed batch with a MatPES TensorNet checkpoint would give us numbers — happy to follow up if useful.

### DIFF
--- a/src/matgl/ext/_ase_dgl.py
+++ b/src/matgl/ext/_ase_dgl.py
@@ -247,24 +247,30 @@ class PESCalculator(Calculator):
             calc_result = self.potential(
                 g=graph, lat=lattice, state_attr=state_attr_default, total_charge=total_charge, ext_pot=self.ext_pot
             )
+
+        # Detach each output tensor to host exactly once per MD step. The
+        # previous code re-ran ``.detach().cpu()`` on the same tensor (energy
+        # twice, stress once or twice with the voigt branch) which forced an
+        # extra GPU→CPU sync per call.
+        host = [t.detach().cpu() for t in calc_result]
+        energy_arr = host[0].numpy()
+        forces_arr = host[1].numpy()
+        energy_val = energy_arr.item()
         self.results.update(
-            energy=calc_result[0].detach().cpu().numpy().item(),
-            free_energy=calc_result[0].detach().cpu().numpy().item(),
-            forces=calc_result[1].detach().cpu().numpy(),
+            energy=energy_val,
+            free_energy=energy_val,
+            forces=forces_arr,
         )
         if self.compute_stress:
-            stresses_np = (
-                full_3x3_to_voigt_6_stress(calc_result[2].detach().cpu().numpy())
-                if self.use_voigt
-                else calc_result[2].detach().cpu().numpy()
-            )
+            stress_arr = host[2].numpy()
+            stresses_np = full_3x3_to_voigt_6_stress(stress_arr) if self.use_voigt else stress_arr
             self.results.update(stress=stresses_np * self.conversion_factor)
         if self.compute_hessian:
-            self.results.update(hessian=calc_result[3].detach().cpu().numpy())
+            self.results.update(hessian=host[3].numpy())
         if self.compute_magmom:
-            self.results.update(magmoms=calc_result[4].detach().cpu().numpy())
+            self.results.update(magmoms=host[4].numpy())
         if self.compute_charge:
-            self.results.update(charges=calc_result[4].detach().cpu().numpy())
+            self.results.update(charges=host[4].numpy())
 
 
 # for backward compatibility
--- a/src/matgl/ext/_ase_pyg.py
+++ b/src/matgl/ext/_ase_pyg.py
@@ -246,22 +246,26 @@ class PESCalculator(Calculator):
             calc_result = self.potential(graph, lattice, self.state_attr)
         else:
             calc_result = self.potential(graph, lattice, state_attr_default)
+
+        # Detach to host once per output tensor — repeated ``.detach().cpu()``
+        # on the same tensor was triggering an extra host sync per MD step.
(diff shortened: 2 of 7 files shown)


shyuep:
### Local PYG benchmark — perf branch vs main (CPU, Apple Silicon)

Microbench at `/tmp/bench_perf.py` (best-of-two median per branch):

| Item | Workload | main (ms) | perf (ms) | Δ |
|---|---|---|---|---|
| 2 — converter | `Structure2Graph.get_graph`, 512 atoms × 80-element table | 0.642 – 0.655 | 0.583 – 0.611 | **−7 to −9%** |
| 3 — loss_fn | `PotentialLightningModule.loss_fn`, batch=64 (E+F+S) | 0.317 – 0.321 | 0.306 – 0.316 | **−2 to −3%** |
| 7 — `PESCalculator.calculate` | 54-atom Li₂O cell, **CPU** | 13.4 – 14.5 | 13.5 – 14.0 | ~−1% (noisy) |

Notes:
- Item 7's gain is muted on CPU because `.cpu()` is a no-op there. The change kills two-to-three redundant GPU→CPU syncs per MD step; the 5–20% number from the audit only materialises on a real CUDA device.
- Item 2 cleanly delivers the predicted ~10% on a worst-case lookup (512 atoms × 80-element table). Real MatPES-style datasets with ~89 elements should be similar or better.
- Item 3's gain matches the small absolute size of three division ops on a six-tensor input; cumulative across thousands of validation steps it adds up.
- Items 4 (mixin grad cleanup) and 6 (DataLoader defaults) only show up in full training runs and weren't covered here. Item 4 affects only `ModelLightningModule` (non-PES). Item 6 only matters when the user wasn't already setting `num_workers`/`pin_memory`.

PR #797, opened 2026-05-28
fix: cast state_attr to long before nn.Embedding lookup in EmbeddingBlock

bowen-bd:
## Summary

- `MEGNet.predict_structure()` constructs `state_attr` with `dtype=matgl.float_th` (float32)
- `EmbeddingBlock.forward()` passed it directly to `nn.Embedding` when `ntypes_state` is set — but `nn.Embedding` requires integer (Long) indices, causing a `FloatTensor` crash for any model using state embeddings (e.g. the BandGap model with PBE/GLLB-SC/HSE/SCAN functionals)
- Fix: add `.long()` cast at the lookup site, consistent with how `ntypes_node` is already handled on the line above

## Minimal reproduction

```python
import matgl
import torch
from pymatgen.core import Structure, Lattice

# MEGNet-BandGap-mfi-MP-2019.4.1 sets ntypes_state=4 (PBE/GLLB-SC/HSE/SCAN)
model = matgl.load_model("MEGNet-BandGap-mfi-MP-2019.4.1")
si = Structure(Lattice.cubic(5.43), ["Si", "Si"], [[0, 0, 0], [0.25, 0.25, 0.25]])

# state_attr dtype=float32, as produced internally by predict_structure → crash
model.predict_structure(si, state_attr=torch.tensor([0], dtype=torch.float32))
# RuntimeError: Expected tensor for argument #1 'indices' to have one of the
# following scalar types: Long, Int; but got torch.FloatTensor instead
```

The one-line fix in `EmbeddingBlock.forward()`:

```python
# before (broken)
state_feat = self.layer_state_embedding(state_attr)
# after
state_feat = self.layer_state_embedding(state_attr.long())
```

## Tests

Two regression tests, both fail on the unfixed code and pass with the fix:

- `tests/layers/test_embedding.py::test_embedding_block_ntypes_state_float_input` — unit test: builds an `EmbeddingBlock` with `ntypes_state=4`, passes a `float32` state index, asserts no crash and correct output shape.
- `tests/models/test_megnet.py::test_predict_structure_bandgap_model` — end-to-end: loads `MEGNet-BandGap-mfi-MP-2019.4.1` from the model hub, calls `predict_structure` with a `float32` state index, asserts a finite scalar result.

## Test plan

- [ ] `pytest tests/layers/test_embedding.py::test_embedding_block_ntypes_state_float_input`
- [ ] `pytest tests/models/test_megnet.py::test_predict_structure_bandgap_model`
- [ ] Verify no regression for models without `ntypes_state` (the cast is inside that branch only)

### DIFF
--- a/src/matgl/layers/_embedding.py
+++ b/src/matgl/layers/_embedding.py
@@ -103,7 +103,7 @@ class EmbeddingBlock(nn.Module):
             edge_feat = edge_attr
         if self.include_state is True:
             if self.ntypes_state and self.dim_state_embedding is not None:
-                state_feat = self.layer_state_embedding(state_attr)
+                state_feat = self.layer_state_embedding(state_attr.long())
             elif self.dim_state_feats is not None:
                 state_attr = torch.unsqueeze(state_attr, 0)
                 state_feat = self.layer_state_embedding(state_attr.to(matgl.float_th))
--- a/tests/layers/test_embedding.py
+++ b/tests/layers/test_embedding.py
@@ -1,12 +1,35 @@
 from __future__ import annotations
 
+import torch
 from torch import nn
 
 import matgl
-from matgl.layers import BondExpansion
+from matgl.layers import BondExpansion, EmbeddingBlock
 from matgl.layers._embedding import (
     TensorEmbedding,
 )
+
+
+def test_embedding_block_ntypes_state_float_input():
+    # Regression test: EmbeddingBlock must accept float32 state_attr when ntypes_state is set.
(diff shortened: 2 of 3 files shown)


shyuep:
Pls add a unittest (using Claude or otherwise). Thanks.

bowen-bd:
The previous megnet tests unfortunately didn't cover the case that requires `ntypes_state`.
So the actual usage of `"MEGNet-BandGap-mfi-MP-2019.4.1"` was not tested and guaranteed.
This should be covered and fixed now

shyuep:
Thanks!

PR #809, opened 2026-06-19
fix(layers): make SoftExponential autograd-correct and NaN-safe (#788)

### LINKED ISSUE #788: Bug: SoftExponential.forward branches on a learnable parameter, breaking autograd and producing NaNs
## Summary

`SoftExponential` in `src/matgl/layers/_activations.py` uses a Python `if`/`elif` ladder on its learnable `alpha` parameter inside `forward`. This has four issues, ranging from a real correctness bug (gradients of `alpha` are computed for only one branch at a time) to numerical and performance footguns.

## Affected code

`src/matgl/layers/_activations.py:62-75`:

```python
class SoftExponential(nn.Module):
    def __init__(self, alpha: float | None = None):
        super().__init__()
        if alpha is None:
            self.alpha = nn.Parameter(torch.tensor(0.0))
        else:
            self.alpha = nn.Parameter(torch.tensor(alpha))
        self.alpha.requires_grad_(True)

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        if self.alpha == 0.0:
            return x
        if self.alpha < 0.0:
            return -torch.log(1.0 - self.alpha * (x + self.alpha)) / self.alpha
        return (torch.exp(self.alpha * x) - 1.0) / self.alpha + self.alpha
```

`self.alpha` is a learnable `nn.Parameter` with `requires_grad=True`.

## Bug 1 — branch decision discards the autograd graph (correctness)

`self.alpha == 0.0` and `self.alpha < 0.0` return a 0-d boolean tensor; the `if` implicitly calls `bool(...)`, which both forces a host-device sync and **drops the comparison from the autograd graph**. Consequences:

- One of the three formulas is picked per call.
- `dL/d(alpha)` is computed only along the branch that fired.
- The decision boundaries (`alpha == 0`, `alpha → 0⁻`, `alpha → 0⁺`) are not differentiable points, and the activation is not piecewise-differentiable in a way autograd can see.
- If `alpha` is currently `−0.01` and the optimum sits at some small `+ε`, the optimizer gets no signal pointing across the boundary. The parameter is effectively trapped in whichever sign region it was initialised in.

A correct piecewise-differentiable implementation uses `torch.where` so both branches' partials in `alpha` participate in the autograd graph. The unused branch must be guarded with safe stand-in inputs (the "double-where trick") because `torch.where` still evaluates both sides:

```python
def forward(self, x):
    a = self.alpha
    safe_log_arg = torch.where(a < 0, 1.0 - a * (x + a), torch.ones_like(x))
    neg = -torch.log(safe_log_arg) / a
    pos = torch.expm1(a * x) / a + a
    out = torch.where(a < 0, neg, pos)
    return torch.where(a.abs() < eps, x, out)
```

## Bug 2 — exact equality test on a float parameter

`self.alpha == 0.0` compares a learned float to exactly zero. After the first optimizer step, `alpha` is almost certainly some tiny non-zero float (e.g. `1e-7`), so the "identity" branch is unreachable in practice. The intent — "behave as identity near zero" — needs a tolerance, e.g. `torch.abs(a) < eps`.

## Bug 3 — undefined-region values silently produce NaN/Inf

For `a < 0`, the formula `−log(1 − a(x + a)) / a` requires `1 − a(x + a) > 0`, i.e. (with `a < 0`) `x > −a − 1/|a|`. For sufficiently negative `x` the `log` argument goes ≤ 0, producing NaN/Inf. In a PES context where features can be large early in training, this is easy to trigger. Fix by clamping the `log` argument to a small positive floor, or reformulating via `softplus`.

## Bug 4 — host-device sync on every call (perf)

`if self.alpha == 0.0:` on a CUDA `alpha` triggers a `.item()` round-trip on every forward, stalling the GPU. `torch.compile` cannot fuse across this. The `torch.where` rewrite removes it.

## Blast radius

`SoftExponential` is not the default for any shipped model. It is selectable via `ActivationFunction` and via string-config model assembly, so a user who picks it hits all four bugs at once. The "softplus2", "swish", "silu" defaults are unaffected.

## Suggested fix

A `torch.where`-based rewrite with the double-where guard plus an `eps` tolerance on the zero-branch, as sketched above. Alternatively, deprecate `SoftExponential` from the `ActivationFunction` enum until fixed.

Happy to send a PR.

0xSoftBoi:
Fixes #788.

`SoftExponential.forward` selected its three formulas with a Python `if` on the learnable `alpha` parameter. This refactors the branch selection to `torch.where` and guards the numerically undefined regions.

## Problems with the old `forward`

```python
if self.alpha == 0.0:
    return x
if self.alpha < 0.0:
    return -torch.log(1.0 - self.alpha * (x + self.alpha)) / self.alpha
return (torch.exp(self.alpha * x) - 1.0) / self.alpha + self.alpha
```

1. **Autograd correctness.** `if self.alpha < 0.0` calls `bool()` on a 0-d tensor, which forces a host-device sync and drops the branch decision from the autograd graph. The activation is not differentiable in a way autograd can see across the sign boundary, so `alpha` is effectively trapped in whichever sign region it was initialised in.
2. **Unreachable identity branch.** `self.alpha == 0.0` is an exact-float comparison on a *learned* parameter. After the first optimizer step `alpha` is some tiny non-zero float, so the intended "identity near zero" behaviour never fires.
3. **NaN/Inf.** For `alpha < 0`, the log argument `1 - alpha*(x + alpha)` goes `<= 0` for sufficiently negative `x`, producing NaN/Inf that then propagates into `alpha.grad` and silently kills training.

## Fix

- Select branches with `torch.where`, keeping both formulas in the autograd graph.
- Treat `|alpha| < eps` as the identity (`alpha -> 0`) region instead of testing exact equality to `0.0`.
- Guard the `log` argument (clamp to a positive floor) and the denominators (the "double where" trick) so both the activation and `alpha.grad` stay finite on the discarded branch too.
- Use `expm1` for the `alpha > 0` branch for accuracy near zero.

Values in the well-defined region are **numerically unchanged** (verified max abs diff `0`–`5e-7` vs the old formula, the residual being `expm1` rounding).

## Tests

Added regression tests in `tests/layers/test_activations.py`:
- output is finite for large-negative `x` with `alpha < 0`,
- `alpha.grad` is finite for those same inputs,
- `alpha` receives a finite, non-zero gradient for both sign initialisations.

These fail on the current code (NaN output / NaN `alpha.grad`) and pass after the fix. Full `tests/layers/` passes (92), `ruff check`/`ruff format` and `mypy` are clean.

One existing assertion in `test_soft_exponential` was updated from `.numpy()` to `.detach().numpy()` for the `alpha == 0` case, because the output now correctly participates in autograd instead of short-circuiting to a detached `x`.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -7,6 +7,14 @@ nav_order: 3
 # Change Log
 
 ## Unreleased
+- **Fix: `SoftExponential` activation autograd correctness and NaN safety (#788).** `forward` now selects
+  its `alpha < 0` / `alpha > 0` / `alpha ≈ 0` branches with `torch.where` instead of a Python `if` on the
+  learnable `alpha` parameter. The old `if self.alpha < 0.0` forced a host-device sync and dropped the
+  branch from the autograd graph (so `alpha` was effectively trapped in its initial sign region); the
+  `alpha == 0.0` exact-float test was unreachable after the first optimizer step; and the `alpha < 0`
+  formula produced NaN/Inf for sufficiently negative inputs. The log argument and denominators are now
+  guarded so both the activation and `alpha.grad` stay finite. Values in the well-defined region are
+  unchanged.
 - **New: `matgl.utils.MCDropoutWrapper` for uncertainty-aware inference.** Enables Monte Carlo Dropout
   (Gal & Ghahramani, 2016) on any pretrained MatGL model (CHGNet, M3GNet, TensorNet, …) without
   retraining: the backbone stays in `eval()` while only the readout dropout is sampled, and
--- a/src/matgl/layers/_activations.py
+++ b/src/matgl/layers/_activations.py
@@ -56,6 +56,11 @@ class SoftExponential(nn.Module):
     References: https://arxiv.org/pdf/1602.01321.pdf
     """
 
+    # |alpha| below this is treated as the identity (alpha -> 0) region, and it
+    # also floors the log argument / denominator to keep the activation and its
+    # gradient finite.
+    _eps = 1e-6
+
     def __init__(self, alpha: float | None = None):
         """Init SoftExponential with alpha value.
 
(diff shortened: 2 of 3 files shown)


kenko911:
@0xSoftBoi Thanks! It looks good to me

PR #815, opened 2026-06-24
Add LAMMPS pair_matgl + Kokkos plugin, add GPU CI job

drakeyu:
## Summary
- Fix the `pair_matgl`/`pair_matgl/kk` ghost-row bug: periodic neighbors were folded onto duplicate ghost rows with `unit_shifts` always zero, which silently dropped TensorNet's outgoing messages since the model needs exactly one consistent row per physical atom. Now every edge resolves to its local row via `atom->map()`, with an explicit integer `unit_shifts` recovered from the ghost/local position offset — fixes a ~30-42 eV energy gap against the native matgl/ASE calculator.
- Sync `lammps/src/KOKKOS/pair_matgl_kokkos.{cpp,h}` with the version actually exercised on hardware (this fix + the virial sign fix + a `Kokkos::LayoutRight` fix for `edge_index`/`unit_shifts` views + correct GPU device selection + a multi-rank guard) — previously these only existed in a separately-patched build tree and had drifted from this repo.
- Update the README's "Implementation notes" to describe the actual algorithm instead of the old (buggy) ghost-position description.
- Add a `build_pair_matgl_kokkos` CI job mirroring the existing CPU job, for when a self-hosted CUDA runner is registered against this repo.

## Test plan
- [x] Single-point energy/stress on the 4-atom Mo-S test deck (`lammps/tests/in.matgl_si`) matches the Python reference (`python_reference.py`) to ~2e-6 eV, run via `-k on g 1 -sf kk` on an A100 GPU.

### DIFF
--- a/.github/workflows/lammps-build.yml
+++ b/.github/workflows/lammps-build.yml
@@ -9,9 +9,12 @@ name: LAMMPS pair_matgl build
 #   * gcc/g++/cmake/python3.
 # If any of those are missing the workflow falls back to downloading them.
 #
-# Kokkos / GPU variant is *not* exercised here — GitHub-hosted runners have
-# no GPU. Phase-3 hardware testing is delegated to a self-hosted CUDA
-# runner (TODO).
+# A second job, build_pair_matgl_kokkos, exercises the Kokkos/GPU variant
+# (pair_style matgl/kk). It needs a self-hosted runner labeled
+# [self-hosted, gpu, cuda] with an NVIDIA driver + nvcc on PATH — there is
+# no such runner registered against this repo today, so the job will queue
+# forever until one exists. Verified manually on a single-GPU NERSC node;
+# not yet exercised by this workflow.
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


shyuep:
🤖 **Automated PR review** (generated by Claude on behalf of @shyuep). Based purely on the diff + current GitHub Actions state; no code was executed locally.

Substantive correctness fix to the `pair_matgl/kk` Kokkos pair style, plus syncing the in-repo source with the hardware-tested build.

**Ghost-row fix (the core):** Edges now resolve to the owning local row via `atom->map(atom->tag[j])`, with an explicit integer `unit_shifts` recovered from `x[j] - x[j_local]` through `domain->h_inv` and rounded. This matches the CPU `pair_matgl` approach and the affine `x2lamda` map (boxlo cancels in the difference — correct). Folding every edge onto one consistent physical-atom row is the right fix for TensorNet message passing, and explains the ~30–42 eV gap. ✅

**Virial sign fix:** `virial -= va` (model returns `dE/dstrain = -W`; LAMMPS convention is `W = Σ r⊗f`), consistent with the `pair_gnnp` reference. Good catch. ✅

**LayoutRight on `edge_index`/`unit_shifts`** ensures contiguous rows matching the torch tensor layout; **GPU selection via `cudaGetDevice()`** is more robust than `local_rank`; **multi-rank guard** (`error->all` when `nprocs>1`) correctly enforces the single-rank assumption and is documented in the README. ✅

**Minor:** `local_row_of_` is rebuilt host-side every `compute()` (an `atom->map` loop + `deep_copy`), i.e. O(nall) host work + a sync per step. Fine for the supported single-GPU small-system case and the comment acknowledges it, but worth revisiting if larger systems become a target.

**⚠️ CI coverage gap:** the only green check here is `lint`. The CPU parity job (`lammps-build`) does not appear to have run on this PR, and the new `build_pair_matgl_kokkos` job targets `[self-hosted, gpu, cuda]` — no such runner is registered, so it will queue until `timeout-minutes: 60`. Net: the fix is currently validated only by the manual A100 run described in the PR body, with no executed CI parity check. Recommend confirming the CPU `lammps-build` parity job runs (and passes) on this PR before merge.

**Verdict:** Code is correct and well-reasoned — LGTM on the diff. Main caveat is the absence of executed CI parity coverage; merge once the CPU job is green or hardware parity is re-confirmed.

shyuep:
@drakeyu Thanks. I merged this version first. You can keep tweaking and submit new PRs as we improve this.

PR #817, opened 2026-06-30
Mkphuthi/training helpers [WIP]

mkphuthi:
## Summary

Major changes:

- feature 1: Added some scripts to train and finetune models with flexibility to change datasets, schedulers and optimizers. 
    - I have found that finetuning can be unstable with the default CosineAnnealingLR scheduler, so this flexibility allows using LR warmup and another scheduler such as ReduceLROnPlateau.  Other best practices can be integrated as we find them
    - The training script is driven entirely by a config yaml/json with documented examples provided
    - The trainers load either pymatgen jsons or ASE-readable files and allows the user to specify the stress unit so conversion is done internally
 
- fix 1: The scheduler was stepping twice per epoch, this is the only change that affects the original source. It does not affect trained models but does point out that every model that has been trained so far has actually been stepped for half as many epochs as the reported number

## Todos

- Right now the training script is run by directly calling them from the scripts directory. Perhaps a cleaner implementation would be to integrate with the cli so you can call "mgl train config.yaml" or "mgl finetune config.yaml" but this requires a more significant refactor

## Checklist

- [ ] Train a FP to test

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -7,6 +7,11 @@ nav_order: 3
 # Change Log
 
 ## Unreleased
+- **Config-driven training scripts (`scripts/train/`).** New `train_pes.py` and `finetune_pes.py`,
+  driven by a single YAML/JSON config (example configs included). Support: optimizer/scheduler by name
+  (`torch.optim` / `lr_scheduler`, default `CosineAnnealingLR`), optional linear LR warmup, datasets
+  from pymatgen JSON or ASE files (`.extxyz`/`.traj`), `stress_unit` conversion to GPa, train/val/test
+  or on-the-fly split, checkpoint/resume, and CSV/W&B logging.
 - **Bug fix: LR schedulers now step once per epoch.** `MatglLightningModuleMixin` previously advanced
   the learning-rate scheduler twice per epoch — once in `on_train_epoch_end` and once via Lightning,
   because `configure_optimizers` returned the `([opt], [sched])` list form that Lightning also steps.
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -6,6 +6,19 @@ nav_order: 3
 
 # Change Log
 
+## Unreleased
+- **Bug fix: LR schedulers now step once per epoch.** `MatglLightningModuleMixin` previously advanced
+  the learning-rate scheduler twice per epoch — once in `on_train_epoch_end` and once via Lightning,
+  because `configure_optimizers` returned the `([opt], [sched])` list form that Lightning also steps.
+  `configure_optimizers` now returns an explicit `lr_scheduler` config (`interval="epoch"`,
+  `frequency=1`, plus `monitor` for `ReduceLROnPlateau`) and the manual `on_train_epoch_end` step has
+  been removed. **This does not affect any saved/deployed model** (the Lightning module is a training
+  harness and is never serialized; `matgl.load_model(...)` and inference are unchanged), but **models
+  retrained with the default `CosineAnnealingLR(T_max=decay_steps)` will differ slightly**, since the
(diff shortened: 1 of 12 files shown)


shyuep:
This is fine, but don't we already have a MGLPotentialTrainer? I suggest we work from that and add the config.yaml and CLI.

shyuep:
🤖 **Automated PR review by Claude** (posted via @shyuep). Based on static analysis of the diff; no code was executed locally. Note: PR is marked **[WIP]**, so this is preliminary.

**Config-driven training scripts + LR-scheduler fix.** Well-structured and thoroughly documented. Main comments:

- ✅ **The LR-scheduler double-step fix is real and correct.** Returning `([opt],[sched])` while also stepping in `on_train_epoch_end` did advance the schedule twice/epoch. Moving to the explicit `lr_scheduler` config (`interval`/`frequency`, `monitor` only for `ReduceLROnPlateau`) is the right fix.
- ⚠️ **Backward-compat / behavior change for all existing users**, not just the new scripts: with the double-step removed, default `CosineAnnealingLR(T_max=decay_steps)` now spans `decay_steps` epochs instead of `decay_steps/2`. Retrained models will differ. The changelog documents this and gives the `decay_steps/2` workaround — good — but please call it out prominently in release notes since it silently changes existing training runs.
- 🔹 Per @shyuep's comment: reconcile with the existing `MGLPotentialTrainer` before adding a parallel training path — worth confirming these scripts build on it rather than duplicating dataset/trainer plumbing.
- 🔹 `lr_scheduler_monitor` default `"val_Total_Loss"` only matters for `ReduceLROnPlateau`; confirm that metric key is actually logged during validation, else plateau scheduling will error.
- 🔹 No CI checks are reported on this branch — get lint + tests green (incl. the new `tests/scripts/train/` and `tests/utils/test_training.py`) before this leaves WIP.

Nice work overall; the scheduler bug fix alone is valuable.

shyuep (commented):
*Automated PR review generated by Claude on behalf of @shyuep.*

Useful direction: config-driven train/finetune scripts with pluggable schedulers/optimizers (LR warmup + ReduceLROnPlateau for finetuning stability) plus tests. Noted as WIP; current blockers:

- Branch has merge conflicts with main — please rebase.
- CI has not run on this branch, so no automated verification yet.
- At +1597/−14 across scripts, `utils/training.py`, and tests, consider whether the `training.py` changes can land separately from the example scripts to ease review.

Happy to re-review once conflicts are resolved and CI runs.

shyuep (commented):
**🤖 Automated PR review generated by Claude** (posted via @shyuep's scheduled review routine)

The most important part of this PR is the scheduler fix in `src/matgl/utils/training.py`: the old code both returned the scheduler from `configure_optimizers` (Lightning auto-steps it per epoch) *and* stepped it manually in `on_train_epoch_end`, so every scheduler was stepped 2x per epoch. Moving to a proper `lr_scheduler` config dict (with `interval`/`frequency`, and `monitor` for `ReduceLROnPlateau`) is the correct fix, and the implication that all previously trained models effectively used half the reported decay schedule is worth documenting prominently.

Comments on the rest:

- The config-driven train/finetune scripts (`scripts/train/`) are useful, and the example YAMLs + tests are appreciated. Agree with the stated TODO that `mgl train config.yaml` CLI integration would be cleaner — fine as a follow-up.
- `lr_scheduler_interval`/`lr_scheduler_monitor` plumbed through both `ModelLightningModule` and `PotentialLightningModule` consistently — good.
- No CI checks have run on this branch (likely awaiting workflow approval as a first-time-contributor branch) — needs a maintainer to approve the run before this can be evaluated for merge. Still marked WIP with an unchecked "train a FP to test" item.

Suggest splitting consideration: the scheduler fix is merge-ready on its own merits once CI runs; the training scripts can iterate.

PR #825, opened 2026-08-20
Fix LAMMPS Kokkos CUDA and CPU virial compatibility

so2koo:
This PR fixes three LAMMPS integration issues affecting current Kokkos/CUDA and CPU builds:

  1. Updates Kokkos coordinate-view access to support current LAMMPS/Kokkos APIs.
  2. Removes illegal implicit `this` captures from CUDA Kokkos kernels.
  3. Corrects the CPU virial sign convention.
  4. Replaces obsolete `ArrayTypes::t_x_array` usage with current LAMMPS-compatible types.

  ## Changes

  - `lammps/src/KOKKOS/pair_matgl_kokkos.cpp`
    - Uses explicit device-captured data inside Kokkos lambdas.
    - Updates coordinate and per-atom view handling.
    - Removes reliance on obsolete `ArrayTypes::t_x_array`.

  - `lammps/src/ML-MATGL/pair_matgl.cpp`
    - Corrects the CPU virial sign to match LAMMPS conventions.

### DIFF
--- a/.github/workflows/lammps-build.yml
+++ b/.github/workflows/lammps-build.yml
@@ -310,13 +310,28 @@ jobs:
               sys.exit(1)
           PY
 
+      - name: Validate pressure sign by finite differences
+        shell: bash
+        env:
+          LIBTORCH: ${{ steps.libtorch.outputs.path }}
+        run: |
+          export LD_LIBRARY_PATH="${LIBTORCH}/lib:${LD_LIBRARY_PATH:-}"
+          cp lammps_artifacts/model.pt matgl/lammps/tests/model.pt
+          cd matgl/lammps/tests
+          LMP="${GITHUB_WORKSPACE}/lammps_build/lmp"
+          "${LMP}" -var scale 0.99 -log log.strain-minus.lammps -in in.matgl_finite_strain
+          "${LMP}" -var scale 1.00 -log log.strain-zero.lammps  -in in.matgl_finite_strain
+          "${LMP}" -var scale 1.01 -log log.strain-plus.lammps  -in in.matgl_finite_strain
+          python3 check_finite_difference_pressure.py \
+            log.strain-minus.lammps log.strain-zero.lammps log.strain-plus.lammps
+
       - name: Upload artifacts on failure
         if: failure()
         uses: actions/upload-artifact@v4
         with:
           name: lammps-debug
           path: |
-            matgl/lammps/tests/log.lammps
+            matgl/lammps/tests/log*.lammps
             lammps_artifacts/
             lammps_build/CMakeFiles/CMakeOutput.log
             lammps_build/CMakeFiles/CMakeError.log
@@ -557,7 +572,9 @@ jobs:
           export LD_LIBRARY_PATH="${LIBTORCH}/lib:${LD_LIBRARY_PATH:-}"
           cp lammps_artifacts/model.pt matgl/lammps/tests/model.pt
           cd matgl/lammps/tests
-          "${GITHUB_WORKSPACE}/lammps_build/lmp" -k on g 1 -sf kk -in in.matgl_si | tee log.lammps
+          "${GITHUB_WORKSPACE}/lammps_build/lmp" \
+            -k on g 1 -sf kk -pk kokkos neigh half newton on \
+            -in in.matgl_si | tee log.lammps
 
       - name: Diff LAMMPS energy against Python reference
         shell: bash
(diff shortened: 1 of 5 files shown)


shyuep:
@drakeyu pls review.

kenko911:
Maybe we should add a unit test to verify that we can reproduce the values obtained with the AdvancedSoft MatGL–LAMMPS interface? It seems we don't have ground truth here.

drakeyu:
Hi,

I have reviewed the code, this PR does change some potential build-compatibility issues and corrects the issue in CPU build that I haven't checked before. The CI failure is unrelated to this PR's diff; the workflow's inline export script imports matgl.apps._pes_pyg and matgl.models._tensornet_pyg. Those _pyg-suffixed private modules were removed from main in #792. I would suggest @so2koo fix the imports in .github/workflows/lammps-build.yml (both occurrences) as part of this PR, so CI can actually exercise these changes.

so2koo:
Thanks for review,

The new commit includes that fix, CI can now exercise the LAMMPS changes.

shyuep:
**Automated PR review generated by Claude** (scheduled routine, posted on behalf of @shyuep)

The Kokkos changes look correct: replacing implicit `this`-member captures with explicit local `const auto` view copies is required for CUDA lambdas, and dropping `ArrayTypes::t_x_array` in favor of `auto` from `k_x.view<DeviceType>()` tracks current LAMMPS APIs. Workflow import-path updates match the current module layout. CI is green (one skipped job).

One thing to verify before merge — the **CPU virial sign flip** in `pair_matgl.cpp`: the comment now asserts the model returns dE/dstrain while LAMMPS wants W = −dE/dstrain. Please confirm (a) this matches the actual TensorNet virial output convention, ideally validated against finite-difference pressure or an NPT sanity run, and (b) the Kokkos path applies the same convention — this diff changes only the CPU side, so if the Kokkos pair style accumulates the virial with the old sign, CPU and GPU builds will now disagree.

shyuep (commented):
**[Automated PR review generated by Claude]** (re-review: new commits since last automated review)

Solid set of LAMMPS integration fixes:

- **Kokkos implicit `this` captures**: hoisting member views into local `const auto` copies before the `KOKKOS_LAMBDA`s is the correct fix — member access inside a by-value device lambda dereferences a host `this` on CUDA. All five kernels now capture device views explicitly.
- **`AT::t_x_array` → `auto`**: appropriate for current LAMMPS/Kokkos API drift.
- **CPU virial sign flip**: the convention change (model returns dE/dstrain; LAMMPS wants W = −dE/dstrain) is behavior-changing for existing users, but the new finite-difference pressure check validates the sign against an independent energy derivative rather than the model's own stress — exactly the right test. Nice addition for both CPU and Kokkos paths.
- CI import paths updated to public module names (`matgl.apps.pes` etc.).

Two minor points:
1. Only `vflag_global` is handled; if per-atom virials (`vflag_atom`) are ever requested (e.g. `compute stress/atom`), they'll silently be zero. Worth a follow-up or an explicit error.
2. The `pair_matgl/kk (GPU, Kokkos)` CI job was still pending at review time — wait for it before merging. All other checks pass.

LGTM pending the GPU job.

shyuep (commented):
**🤖 Automated PR review generated by Claude** (posted via @shyuep's scheduled review routine — new commits since last automated review)

Substantive fixes, all in the right direction:

1. **CPU virial sign flip** (`pair_matgl.cpp`): the model outputs dE/dstrain while LAMMPS expects W = −dE/dstrain, so the sign correction is right — and it's a behavior change that silently affected any NPT/pressure users of previous builds. The new finite-difference pressure check in CI (comparing −dE/dλ/(3V) against reported Press for both CPU and Kokkos paths) is exactly the right guard against regression.
2. **Kokkos implicit-`this` captures**: hoisting member views into local `const auto` copies before `KOKKOS_LAMBDA` is the correct idiom for CUDA; the previous code was UB on device.
3. Replacing `ArrayTypes::t_x_array` with `auto` from `k_x.view<DeviceType>()` keeps compatibility with current LAMMPS.

One blocker: the `pair_matgl/kk (GPU, Kokkos)` job failed at the 24h limit — that looks like a hung/stalled GPU runner rather than a test failure, but it needs a rerun (or investigation of a device-side hang, e.g. in the new `-pk kokkos neigh half newton on` path) before merge. Everything else is green.

toohardtogetname:
Independent corroboration of the CPU virial sign fix, from a separate audit of `pair_matgl` (now #828, which deliberately leaves the virial change to this PR):

We hit the same defect and verified it convention-free by finite differences on the 4-atom Mo–S cell of `lammps/tests/in.matgl_si` at `run 0` (no kinetic term), comparing three independent numbers:

| quantity | mean pressure (bar) |
|---|---|
| finite-difference dE/dV | **−88488.5** |
| ASE-convention stress via P = −tr(σ)/3 (`python_reference.py`) | **−88496.4** |
| unpatched `pair_matgl` (`virial += va`) | **+88496.3** |

Right magnitude, inverted sign — exactly what `virial -= va` fixes, and consistent with the model returning dE/dstrain while LAMMPS expects W = −dE/dstrain. Under NPT the unpatched sign drives the barostat the wrong way (the box runs away), so this is worth landing ahead of any NPT use of the CPU style.

The Kokkos style has carried the correct `-=` since #815, so after this PR the two styles agree.

PR #828, opened 2026-08-25
pair_matgl follow-up: style registration, GPU export, ghost-row trimming (3x node work), docs

toohardtogetname:
Follow-up to #815 (`Add LAMMPS pair_matgl + Kokkos plugin`). Building that
plugin from a clean tree exposed five defects: two that make it unusable
(the styles never register; the exported model cannot run on a GPU), one that
silently produces wrong pressures (the CPU style's virial sign — see the #825
note directly below), one that wastes most of the model's node work, and one
broken helper script. Two of the five are invisible to the current CI job,
which compares energy only.

Everything below was reproduced on a clean build and is backed by measurement,
not inspection alone.

## Overlap with #825 — read this first

This work was done independently of and concurrently with #825
(*Fix LAMMPS Kokkos CUDA and CPU virial compatibility*). The two PRs meet at
exactly one point and are otherwise disjoint:

- **CPU virial sign: same finding, deferred to #825.** We found the identical
  defect (the model returns dE/dstrain; LAMMPS' `virial` is W = −dE/dstrain,
  so `+=` must be `-=`) and verified it convention-free by finite differences:
  on the 4-atom Mo–S cell of `lammps/tests/in.matgl_si` at `run 0`,
  dE/dV gives a mean pressure of **−88488.5 bar**, the ASE-convention stress
  gives **−88496.4 bar**, and the unpatched code reported **+88496.3 bar** —
  right magnitude, inverted sign, which under NPT drives the barostat the
  wrong way. These numbers independently corroborate #825's fix and its
  finite-difference CI check; this PR **no longer touches the virial code**
  and defers that change entirely to #825.
- **Everything else is disjoint.** #825's lambda-capture legalization,
  coordinate-view API updates and workflow-import fixes do not intersect the
  items below; conversely #825 does not touch style registration, the
  exported model's GPU placement, ghost-row trimming, or the export script.
- **Merge-order note.** #825's member-view → local-capture rewrite in
  `pair_matgl_kokkos.cpp` (its lines 133–301) textually overlaps item 5's
  hunks here (lines 175–381). This PR should land **after** #825; item 5's
  patch will be rebased onto the post-#825 tree, and its lambda bodies
  switched to #825's captured locals rather than member views.

**Environment.** LAMMPS `stable_22Jul2025_update5`; libtorch 2.7.1+cu128;
matgl 4.0.3 (`25b3a29`); NVIDIA H200 (sm_90), CUDA 12.9, gcc 11.5;
`materialyze/TensorNet-PES-MatPES-r2SCAN-2025.2` exported at float32.

---

## 1. The documented build cannot register the pair styles

**Symptom.** Following `lammps/README.md` exactly produces a binary that
compiles and links libtorch, but rejects the style at run time:

```
ERROR: Unrecognized pair style 'matgl' (src/force.cpp:275)
Last input line: pair_style matgl
```

**Cause.** The README instructs:

```bash
echo 'include(/path/to/matgl/lammps/cmake/ML-MATGL.cmake)' >> <lammps>/cmake/CMakeLists.txt
```

LAMMPS builds its style tables by scanning package directories and then
generating headers. In `cmake/CMakeLists.txt` the ordering is

| line | statement |
|---|---|
| 664 | `include(StyleHeaderUtils)` |
| 665 | `RegisterStyles(${LAMMPS_SOURCE_DIR})` |
| 681–699 | per-package loop: `RegisterStyles` + `target_sources` + include dir |
| 764 | `foreach(PKG_WITH_INCL … KOKKOS …)` → pulls in `Packages/KOKKOS.cmake` |
| **795** | **`GenerateStyleHeaders(${LAMMPS_STYLE_HEADERS_DIR})`** |
| ~1176 | ← where appending puts the `include()` (end of file) |

So the snippet is evaluated ~380 lines *after* the style headers have already
been written. `ML-MATGL.cmake` also never calls `RegisterStyles`, so even
including it early would not register anything: it only does `target_sources`
and links Torch. The result is a pair style that exists in the binary but not
in `style_pair.h`.

**Fix in this PR.** Register the styles the way LAMMPS itself does:

* `ML-MATGL` is added to `STANDARD_PACKAGES`, so the normal package loop does
  `RegisterStyles` + `target_sources` + `target_include_directories` at the
  right point. `-D PKG_ML-MATGL=ON` keeps working unchanged.
* `pair_matgl_kokkos.{cpp,h}` are installed into `src/KOKKOS/`, where
  `Packages/KOKKOS.cmake:219`'s
  `RegisterStylesExt(${KOKKOS_PKG_SOURCES_DIR} kokkos KOKKOS_PKG_SOURCES)`
  picks up `matgl/kk` automatically and appends the source to the Kokkos
  target. This replaces the separate `src/KOKKOS-MATGL/` layout, which the
  KOKKOS package does not scan.
* What remains of the snippets is a link-only fragment (`find_package(Torch)`,
  `target_link_libraries`, `TORCH_CXX_FLAGS`). Keeping the old
  `target_sources` call as well would compile each translation unit twice
  under two different paths and break the link.

Verified on a pristine `stable_22Jul2025` tree, both sides: following the
OLD README verbatim, the snippet executes (libtorch links) yet
`style_pair.h` carries **zero** matgl entries; following the revised
procedure, `style_pair.h` carries both entries, the build links cleanly,
`lmp -h` lists both styles, and a 300-step run reproduces the baseline
timing.

```
-- Enabled packages: KOKKOS;ML-MATGL
#include "pair_matgl.h"
#include "pair_matgl_kokkos.h"
matgl
matgl/kk
```

## 2. Virial sign — see "Overlap with #825" above

Found independently, verified by finite differences (numbers quoted at the
top), and deferred entirely to #825. No virial change ships in this PR.

## 3. `python_reference.py` invites exactly the wrong conclusion

The helper prints `stress_GPa` (ASE convention, `σ = (1/V) ∂E/∂ε`) directly
beside the LAMMPS `Pxx…` thermo columns, with no note that
`P_LAMMPS = −σ_ASE`. Comparing the two columns as printed makes the
sign-flipped CPU value look like agreement and the correct Kokkos value look
like a bug — which is how this defect survived. This PR prints both
conventions explicitly and states the relation.

## 4. The exported model cannot run on a GPU (`torch.arange` on the default device)

**Symptom.** `pair_style matgl/kk` aborts on the first force evaluation:

```
RuntimeError: Expected all tensors to be on the same device, but found at
least two devices, cuda:0 and cpu!   (pair_matgl_kokkos.cpp:364)
```

**Cause.** `matgl/layers/_basis.py::spherical_bessel_smooth` builds two
constants without a device:

```python
n  = torch.arange(max_n).type(dtype=matgl.float_th)[None, :]   # line 361
dn = [torch.tensor(1.0)]                                       # line 374
```

In Python this is masked by matgl's global default device. Under LibTorch the
default device is CPU, so `n` lands on the host while `r` is on `cuda:0`. The
implication is that the exported TorchScript module had never been executed on
a GPU — i.e. the Kokkos path was never run end to end.

**Fix.** Derive the device from the input: `torch.arange(max_n, device=r.device)`
and `torch.tensor(1.0, device=r.device)`. Device-agnostic, and the Python path
is unchanged.

## 5. Ghost rows are fed to the model although no edge references them

Builds directly on #815's ghost-edge folding: since #815, every edge is
resolved to its owning local row (with explicit unit shifts), so ghost ROWS in
the node buffers are guaranteed-isolated nodes -- no incoming message, masked
out of the energy, exactly zero force. This item removes the wasted node work
that folding made removable; it is a performance follow-up, not a correctness
overlap.

**Performance, not correctness** — but it is most of the cost. #815 correctly
folds every edge back onto a local row (`atom->map(atom->tag[j])`) and carries
periodicity in `unit_shifts`, which is what makes message passing right. After
that fold, `i < nlocal` (the loop bound is `list->inum`, local only) and
`j_local < nlocal` by construction, so **no edge can reference a ghost row any
more**. Yet the tensors handed to the model are still sized `nall`:

```cpp
const int nall = nlocal + atom->nghost;
torch::Tensor positions      = torch::empty({nall, 3}, opts_real);
torch::Tensor atomic_numbers = torch::empty({nall}, opts_long);
torch::Tensor local_or_ghost = torch::empty({nall}, opts_bool);
```

Every one of the `nall − nlocal` ghost rows is therefore an **isolated node**:
it receives no messages, contributes no energy (it is masked out of the sum),
and gets zero force — but the model still computes embeddings, norms and
readout for it. The waste is the ghost fraction, which is set by cell size:

| cell | Nlocal | Nghost | nall/nlocal |
|---|---|---|---|
| 32 Å (1,248 atoms) | 1,248 | 2,502 | **3.00×** |
| 64 Å (9,984 atoms) | 9,984 | 8,174 | 1.82× |
| 96 Å (33,696 atoms) | 33,696 | 17,164 | 1.51× |

**Fix.** Size the model's inputs to `nlocal` in both styles. `local_or_ghost`
stays in the signature and becomes all-true. In the Kokkos variant this remains
a zero-copy view: LAMMPS keeps owned atoms in rows `[0, nlocal)` of `x`, so the
`from_blob` view simply narrows. Force scatter-back loops to `nlocal`; the ghost
entries were exactly zero (no edge touches them), so nothing is dropped.
`local_row_of_` stays sized to `nall` — it is the ghost→owner map that makes the
fold possible.

**Measured.** Results are unchanged and the step gets substantially cheaper.
4-atom Mo–S single point, before vs after:

| | PotEng (eV) | mean pressure (bar) |
|---|---|---|
| reference | −64.972801 | −88 496.4 |
| `matgl` after | −64.972801 | −88 496.396 |
| `matgl/kk` after | −64.972801 | −88 496.386 |

1,248-atom Li₃PO₄–TaCl₅, 1,000 steady-state NVT steps. All four legs were run
in ONE job on ONE GPU, interleaved over two rounds, with decks that differ only
where `pair_matgl` requires (`atom_modify map yes`, `newton on`). Same-job is
essential here: the `gnnp` legs vary by **25%** across nodes on this cluster
(23.0-28.19 ms/step over 8 measurements on 5 nodes), far more than the effect
being claimed, so a cross-job comparison could not have settled it:

| leg | ms/step | round spread |
|---|---|---|
| `pair_style gnnp` + `matgl-warp` (fp32) | 26.98 | 0.9% |
| `pair_style gnnp` + plain PyG (fp32) | 33.94 | 2.6% |
| `pair_style matgl/kk`, before this fix | 43.56 | 0.3% |
| **`pair_style matgl/kk`, after this fix** | **25.47** | 0.2% |

so this fix is worth **1.71×** on that cell. That figure is the same binary
before and after, in one job on one GPU, so node variation cannot touch it; the
same holds for the **1.33×** against `gnnp`+PyG, the backend the plugin actually
shares.

Against `gnnp`+`matgl-warp` the ordering is **not stable enough to state as a
ranking, and this PR does not claim one**. On the node above the plugin lands 6%
ahead (25.47 vs 26.98); on other nodes it lands behind. The sensitivity is
asymmetric: `gnnp`+`matgl-warp` is bound by Python-side dispatch and spans 25%
across nodes, while `pair_style matgl/kk` is a C++ style whose round-to-round
spread is 0.2%. At this cell size the two are close enough that which one "wins"
is a property of the node, not of the code.

The gain tracks the ghost fraction, so it is largest on small cells — and it is
smaller than the ghost ratio itself because only the NODE work shrinks, not the
edge work:

| cell | nall/nlocal | before | after | gain |
|---|---|---|---|---|
| 1,248 atoms (32 Å) | 3.00 | 43.56 | **25.47** | **1.71×** |
| 9,984 atoms (64 Å) | 1.82 | 241.06 | **165.68** | 1.46× |
| 33,696 atoms (96 Å) | 1.51 | 750.89 | **543.16** | 1.38× |

Against the same backend the plugin is now consistently ahead — `matgl/kk` vs
`gnnp`+PyG is 1.33× at 1,248 atoms and 1.35× at 9,984 (165.68 vs 224.33). The
two larger sizes are cross-job, so they carry the 25% node caveat; both gaps are
comfortably outside it, unlike the 6% one discussed above. It still trails
`gnnp`+`matgl-warp` at the larger sizes (165.68 vs 97.83 ms at 9,984, i.e. 1.69×
— also well outside the node spread) purely because that backend is unavailable
to the plugin,
and warp's advantage over PyG grows with size (1.24× at 1,248, 2.29× at 9,984).
Making the exporter support `use_warp=True` would be the natural follow-up, and
is out of scope here.

(Note that `pair_matgl` is PyG-only, so it cannot use the `matgl-warp`
backend that `pair_style gnnp` can — the numbers above are with the plugin on
the slower of the two matgl backends.)

## 6. `export_matgl_checkpoint.py` imports modules that no longer exist

`pair_matgl`'s `coeff()` shells out to this script when `pair_coeff` is given a
checkpoint directory, but the script imports the pre-4.0.3 module layout:

```python
import matgl.graph._compute_pyg as _cpyg          # now matgl.graph._compute
from matgl.ext._lammps import export_lammps_model # now matgl.ext.lammps
```

The first raises `ModuleNotFoundError` on 4.0.3, so the "point `pair_coeff` at
a checkpoint dir" path is dead. (`matgl.ext._lammps` does still import on 4.0.3
and still exports `export_lammps_model`, but the script never gets that far.)
Updated to `matgl.graph._compute` / `matgl.ext.lammps`; the
`create_line_graph_torch` shim is no longer needed because `_compute` exports it
directly.

## 7. Documentation gaps that block a first run

* **Kokkos requires `neigh half`.** `pair_matgl` errors out unless
  `newton on`, and LAMMPS refuses `newton on` together with the Kokkos default
  `neigh full`:

  ```
  ERROR: Must use 'newton off' with KOKKOS package option 'neigh full'
  (src/KOKKOS/kokkos.cpp:693)
  ```

  So every Kokkos run needs `-pk kokkos neigh half` (or `package kokkos neigh
  half` before `atom_style`). The README's run line omits this and fails as
  printed.

* **libtorch must be a CUDA build for the Kokkos variant.** The README says
  "LibTorch 2.2.x – 2.5.x (CXX11 ABI, CPU build)", but
  `pair_matgl_kokkos.cpp` selects `torch::Device(torch::kCUDA, gpu)` and wraps
  Kokkos device buffers as tensors without a copy. A CPU-only libtorch runs the
  model on the host. The CPU-build note applies to the CI job only.

* **Two CMake patches are needed for CUDA ≥ 12.9 toolkits.** libtorch's Caffe2
  config predates both, and each aborts the generate step. CUDA 12.9 removed
  the nvToolsExt shared library, so `FindCUDAToolkit` no longer defines
  `CUDA::nvToolsExt` while `Caffe2/public/cuda.cmake:186` still links it into
  `torch::nvtoolsext`; and on a system without MKL, Caffe2 leaves the literal
  `MKL_INCLUDE_DIR-NOTFOUND` inside torch's `INTERFACE_INCLUDE_DIRECTORIES`.
  Declaring `CUDA::nvToolsExt` as a header-only INTERFACE target (nvtx3 headers
  are still shipped) and pointing `MKL_INCLUDE_DIR` at an existing directory
  clears both. Added to the README as a troubleshooting note.

---

## Testing

* **Style registration** — `style_pair.h` contains both headers; `lmp -h`
  lists `matgl` and `matgl/kk`.
* **Single-point parity**, 4-atom Mo–S cell, against
  `python_reference.py` (matgl 4.0.3 / torch 2.7.1):

  | | PotEng (eV) |
  |---|---|
  | python reference | −64.9728012 |
  | `pair_style matgl` | −64.972801 |
  | `pair_style matgl/kk` | −64.972801 |

  Energies agree to printed precision on both styles. Pressures agree in
  both conventions once #825's CPU virial fix is applied (measured with it
  in place; this PR itself does not touch the virial).
* **1,248-atom NVT**, 300 warm-up + 1,000 timed steps, both styles, no NaN,
  temperature in band.
* **GPU execution** — `matgl/kk` reports `model on cuda` and completes
  1,300 steps after fix #4 (before it aborted at step 0).

## Notes on CI

The existing job cannot catch #2 or #4: it diffs energy only, and
GitHub-hosted runners have no GPU. Two cheap additions would close both gaps:

1. the **virial/pressure** gap is already being closed by #825, which adds a
   finite-difference pressure validation to CI — once it lands, #3's `P = −σ`
   labeling makes that check unambiguous to read;
2. run `spherical_bessel_smooth` once under `torch.set_default_device('cpu')`
   with a CUDA input in a unit test — this reproduces #4 without a GPU-enabled
   LAMMPS build.

### DIFF
--- a/lammps/README.md
+++ b/lammps/README.md
@@ -78,6 +78,19 @@ cmake -B build -S <lammps>/cmake \
 cmake --build build -j 8
 ```
 
+Steps 1–3 as a copy-paste block (GNU sed; the `set(STANDARD_PACKAGES`
+opener has kept this exact form across recent LAMMPS releases, and the
+list is not order-sensitive):
+
+```bash
+MATGL=/path/to/matgl
+LMP=/path/to/lammps
+ln -s "$MATGL/lammps/src/ML-MATGL" "$LMP/src/ML-MATGL"
+sed -i '/^set(STANDARD_PACKAGES$/a\  ML-MATGL' "$LMP/cmake/CMakeLists.txt"
+echo "include($MATGL/lammps/cmake/ML-MATGL.cmake)" >> "$LMP/cmake/CMakeLists.txt"
+grep -c ML-MATGL "$LMP/cmake/CMakeLists.txt"   # expect 2: package list + include
+```
+
 Check the registration before running anything:
 
 ```bash
--- a/lammps/README.md
+++ b/lammps/README.md
@@ -38,23 +38,51 @@ of which you'll need for `pair_coeff`.
 
 ### 2. Build LAMMPS with the package
 
-Drop the package into a stock LAMMPS source tree and configure:
+LAMMPS builds its style tables by scanning package directories and then
+generating `style_pair.h`, and the generation happens roughly two thirds of
+the way through `cmake/CMakeLists.txt` (`GenerateStyleHeaders(...)`, line 794
+in `stable_22Jul2025_update5`). **Appending an `include()` to the END of that
+file is therefore too late**: the sources compile and libtorch links, but the
+style never reaches `style_pair.h` and LAMMPS rejects it at run time with
+
+```
+ERROR: Unrecognized pair style 'matgl' (src/force.cpp:275)
+```
+
+Register it the way LAMMPS registers its own packages instead:
 
 ```bash
-# 1) Copy or symlink the source files.
+# 1) Copy or symlink the source files into the LAMMPS src tree.
 ln -s /path/to/matgl/lammps/src/ML-MATGL <lammps>/src/ML-MATGL
 
-# 2) Tell LAMMPS' CMake about the package.
+# 2) Add ML-MATGL to the package list, so LAMMPS' own per-package loop does
+#    RegisterStyles + target_sources + include dir at the right point.
+#    In <lammps>/cmake/CMakeLists.txt, inside set(STANDARD_PACKAGES ...):
+#        ML-IAP
+#      + ML-MATGL
+#        ML-PACE
+#    (`-D PKG_ML-MATGL=ON` then works like any other package flag.)
+
+# 3) Link libtorch. This snippet only does find_package(Torch) and the link;
+#    it must NOT also add the sources, or every file compiles twice under two
+#    paths and the link fails on duplicate symbols.
 echo 'include(/path/to/matgl/lammps/cmake/ML-MATGL.cmake)' \
     >> <lammps>/cmake/CMakeLists.txt
 
-# 3) Configure + build. Match libtorch's CXX11 ABI to LAMMPS'.
+# 4) Configure + build. Match libtorch's CXX11 ABI to LAMMPS'.
 cmake -B build -S <lammps>/cmake \
     -D PKG_ML-MATGL=ON \
     -D CMAKE_PREFIX_PATH=/path/to/libtorch \
     -D CMAKE_BUILD_TYPE=Release \
     -D BUILD_MPI=ON
 cmake --build build -j 8
+```
+
+Check the registration before running anything:
+
+```bash
+grep matgl build/styles/style_pair.h     # expect pair_matgl.h (and _kokkos.h)
+build/lmp -h | tr ' ' '\n' | grep '^matgl'
 ```
 
 ### 2b. Build the Kokkos GPU variant
@@ -64,6 +92,12 @@ matching snippet to LAMMPS' CMake. CUDA example for an Ampere card
 (A100/A30):
 
 ```bash
+# Put the Kokkos sources where the KOKKOS package looks for them: its
+# RegisterStylesExt(${KOKKOS_PKG_SOURCES_DIR} kokkos ...) scans
+# <lammps>/src/KOKKOS for *_kokkos.h style headers and picks up matgl/kk
+# automatically. A separate directory is not scanned.
+cp /path/to/matgl/lammps/src/KOKKOS/pair_matgl_kokkos.* <lammps>/src/KOKKOS/
+
 echo 'include(/path/to/matgl/lammps/cmake/ML-MATGL-KOKKOS.cmake)' \
     >> <lammps>/cmake/CMakeLists.txt
 
@@ -81,8 +115,20 @@ cmake --build build -j 8
 Run with:
 
 ```bash
-mpirun -n 1 build/lmp -k on g 1 -sf kk -in in.matgl_si
+mpirun -n 1 build/lmp -k on g 1 -sf kk -pk kokkos neigh half -in in.matgl_si
 ```
+
+**`neigh half` is required, not optional.** `pair_matgl` needs `newton on`
+(it folds periodic edges back onto local rows and needs ghost contributions),
+and LAMMPS refuses `newton on` together with the Kokkos default `neigh full`:
+
+```
+ERROR: Must use 'newton off' with KOKKOS package option 'neigh full'
+(src/KOKKOS/kokkos.cpp:693)
+```
+
+Equivalently, put `package kokkos neigh half` in the input deck before
+`atom_style`.
 
 `-sf kk` makes LAMMPS prefer Kokkos pair styles, so `pair_style matgl`
 in your input deck dispatches to `matgl/kk` automatically. If you'd
@@ -94,10 +140,30 @@ this explicit.
 
 Tested with:
 
-- LibTorch 2.2.x – 2.5.x (CXX11 ABI, CPU build).
+- LibTorch 2.2.x – 2.7.x (CXX11 ABI). **The Kokkos variant needs a CUDA
+  build of libtorch**, not a CPU-only one: `pair_matgl_kokkos.cpp` selects
+  `torch::Device(torch::kCUDA, gpu)` and wraps Kokkos device buffers as
+  tensors without a copy, so a CPU-only libtorch silently runs the model on
+  the host. The CPU build is what the CI job uses for the serial style.
 - LAMMPS develop branch (Aug 2024 or newer for the `add_request` /
-  `REQ_GHOST` neighbor-list API).
+  `REQ_GHOST` neighbor-list API); verified on `stable_22Jul2025_update5`.
 - C++17, MPI optional.
+
+#### Troubleshooting: CUDA 12.9 and newer toolkits
+
+libtorch's bundled Caffe2 CMake config predates two changes and each aborts
+the generate step. Both are fixed by a small file injected before
+`find_package(Torch)`, e.g. via
+`-D CMAKE_PROJECT_lammps_INCLUDE=/path/to/fixups.cmake`:
+
+- CUDA 12.9 removed the nvToolsExt shared library, so `FindCUDAToolkit` no
+  longer defines `CUDA::nvToolsExt` while `Caffe2/public/cuda.cmake` still
+  links it into `torch::nvtoolsext`. Declare it as a header-only interface
+  target (the nvtx3 headers are still shipped):
+  `add_library(CUDA::nvToolsExt INTERFACE IMPORTED GLOBAL)`.
+- On a machine without MKL, Caffe2 leaves the literal
+  `MKL_INCLUDE_DIR-NOTFOUND` inside torch's `INTERFACE_INCLUDE_DIRECTORIES`.
+  Point `MKL_INCLUDE_DIR` at any existing directory.
 
 ## LAMMPS input syntax
 
--- a/lammps/cmake/ML-MATGL-KOKKOS.cmake
+++ b/lammps/cmake/ML-MATGL-KOKKOS.cmake
@@ -1,43 +1,20 @@
-# ML-MATGL Kokkos variant — drop-in CMake snippet.
+# ML-MATGL Kokkos variant -- documentation / warning fragment.
 #
-# Layered on top of ML-MATGL.cmake: include() this *after* the base snippet,
-# OR set PKG_ML-MATGL=ON and PKG_KOKKOS=ON together.
+# The Kokkos sources are picked up by LAMMPS' own KOKKOS package machinery
+# (Packages/KOKKOS.cmake scans <lammps>/src/KOKKOS for *_kokkos.* styles and
+# registers them via RegisterStylesExt), so this snippet adds NO sources:
 #
-# Usage (from a stock LAMMPS source tree):
-#   cmake -B build \
-#       -D PKG_ML-MATGL=ON -D PKG_KOKKOS=ON \
-#       -D Kokkos_ENABLE_CUDA=ON \
-#       -D Kokkos_ARCH_AMPERE80=ON \
-#       -D CMAKE_PREFIX_PATH=/path/to/libtorch \
-#       -D CMAKE_CXX_COMPILER=$LAMMPS/lib/kokkos/bin/nvcc_wrapper \
-#       <other flags>
+#   cp /path/to/matgl/lammps/src/KOKKOS/pair_matgl_kokkos.* <lammps>/src/KOKKOS/
 #
-# The `pair_matgl/kk` style is registered via the standard LAMMPS Kokkos
-# pair-style macro so users invoke it with `pair_style matgl/kk` or by
-# launching LAMMPS with `-sf kk -k on g 1`.
+# and configure with -D PKG_ML-MATGL=ON -D PKG_KOKKOS=ON (plus the usual
+# Kokkos CUDA flags). libtorch linkage comes from ML-MATGL.cmake.
 
 if(NOT PKG_ML-MATGL OR NOT PKG_KOKKOS)
     return()
 endif()
 
-if(NOT DEFINED ML_MATGL_KOKKOS_DIR)
-    get_filename_component(ML_MATGL_KOKKOS_DIR
-        "${CMAKE_CURRENT_LIST_DIR}/../src/KOKKOS" ABSOLUTE)
-endif()
-
-if(NOT EXISTS "${ML_MATGL_KOKKOS_DIR}/pair_matgl_kokkos.cpp")
-    message(FATAL_ERROR
-        "ML-MATGL-KOKKOS source not found at ${ML_MATGL_KOKKOS_DIR}. "
-        "Set -DML_MATGL_KOKKOS_DIR=<path/to/lammps/src/KOKKOS>.")
-endif()
-
-file(GLOB ML_MATGL_KOKKOS_SOURCES "${ML_MATGL_KOKKOS_DIR}/*.cpp")
-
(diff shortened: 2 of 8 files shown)


shyuep (commented):
**Automated PR review generated by Claude (posted on behalf of @shyuep).**

High-quality follow-up — each fix is measured rather than asserted, and deferring the virial sign to #825 with independent finite-difference corroboration is exactly the right call. CI is green across lint/tests/pair_matgl (CPU); the Kokkos GPU check shows no status yet — confirm it reports before merge.

Comments:

1. **Ghost-row trimming (CPU + Kokkos)**: narrowing positions/atomic_numbers/forces to `nlocal` is correct given the stated invariants (edges folded onto owned rows, single-rank enforced, local atoms in rows `[0, nlocal)`). Since `local_or_ghost` is now all-true, consider dropping the mask tensor from the model interface in a later PR rather than shipping a constant.
2. **CMake restructure**: moving style registration into `STANDARD_PACKAGES` and reducing the snippets to libtorch-link-only fragments is the correct fix for the never-registered styles and the duplicate-symbol hazard. The README's `grep style_pair.h` verification step is a nice guard. One gap: the install is now a 4-step manual edit of LAMMPS' CMakeLists — a small patch script or `sed` one-liner in the README would reduce user error.
3. **`_basis.py` device fixes**: correct and minimal (`torch.arange(..., device=r.device)`, `torch.tensor(1.0, device=r.device)`). This touches core matgl, not just the LAMMPS plugin — the existing unit tests cover CPU; the GPU path is only exercised by the Kokkos CI job, so item on the green GPU check above applies here too.
4. `export_matgl_checkpoint.py` import fix pins to the matgl 4.0.3 layout; fine, but a try/except fallback for older layouts would keep the script usable across the versions the README claims to support.

Recommendation: approve once the Kokkos GPU check reports green.

toohardtogetname:
Thanks for the review — all four items addressed or acknowledged, changes pushed as one commit:

1. **`local_or_ghost` mask**: agreed it is now a constant on this path; leaving the interface change to a follow-up PR as you suggest, since the exported-model signature is shared with existing `.pt` artifacts in the wild.

2. **Scripted install**: the README now carries steps 1–3 as a copy-paste block (symlink + `sed` insertion into `set(STANDARD_PACKAGES` + `include()` append, with a `grep -c` self-check). Verified end-to-end on a pristine `stable_22Jul2025_update5` tree: the block, then configure, and `style_pair.h` carries the pair style. The `set(STANDARD_PACKAGES` opener the sed keys on has kept that exact form across recent LAMMPS releases, and the package list is not order-sensitive.

3. **`_basis.py` GPU coverage**: nothing further from our side — the device fixes ride on the same pending self-hosted Kokkos GPU check. For what it's worth, the fixed file is exercised on GPU daily in our production MD stack.

4. **Exporter fallback**: `export_matgl_checkpoint.py` now tries the 4.0.3 layout (`matgl.ext.lammps`) and falls back to the pre-4.0.3 layout with the `create_line_graph_torch` shim. The fallback branch is verbatim the code this PR previously replaced (i.e., the previously-shipping behavior), and the new-layout branch is verified against matgl 4.0.3.

On the Kokkos GPU check: still shows no status — same behavior as #825's run before it was picked up. Happy to re-trigger if the runner needs a nudge.

shyuep (commented):
**Automated PR review generated by Claude (posted on behalf of @shyuep).**

Follow-up on commit 4eaf45e, which addresses both actionable items from the previous automated review:

1. **README scripted registration block**: the symlink + `sed` into `STANDARD_PACKAGES` + append-include copy-paste block resolves the manual 4-step-edit concern, and the `grep -c ML-MATGL` / `style_pair.h` verification steps guard against silent misregistration. Verified end-to-end on a pristine `stable_22Jul2025_update5` tree per the commit message.
2. **Exporter import fallback**: try matgl ≥4.0.3 layout (`matgl.ext.lammps`) first, fall back to the pre-4.0.3 layout with the `create_line_graph_torch` shim — kept verbatim from the previously replaced code. Resolves the cross-version usability gap.

No new concerns; no source changes beyond the two files above.

**Remaining blocker**: the `pair_matgl/kk (GPU, Kokkos)` check has been QUEUED since 2026-08-26 08:00 UTC (>24 h) — the self-hosted GPU runner is likely offline. This is the only check standing between this PR and the previously recommended approval, and it is the one that exercises the `_basis.py` device fixes and the Kokkos ghost-row trimming. Please check the runner; approve once it reports green.

shyuep:
**Automated PR review** (generated by Claude on behalf of @shyuep).

Exemplary follow-up — clean-tree reproduction, finite-difference verification of the virial finding, and explicit disjointness analysis vs. #825. The diagnosis of the style-registration failure (README `include()` lands ~380 lines after `GenerateStyleHeaders`, and `ML-MATGL.cmake` never called `RegisterStyles`) is convincing, and routing registration through `STANDARD_PACKAGES` + installing the Kokkos sources into `src/KOKKOS/` for `RegisterStylesExt` pickup is the way LAMMPS itself expects packages to integrate.

Action items:

1. **#825 is now merged — please rebase.** Per your own merge-order note, item 5's hunks in `pair_matgl_kokkos.cpp` need to be rebased onto #825's local-capture rewrite.
2. **`_basis.py` device fix is correct.** Minor: fold dtype into construction — `torch.arange(max_n, dtype=matgl.float_th, device=r.device)` — rather than `.type()` after, which allocates twice; same for `torch.tensor(1.0, dtype=matgl.float_th, device=r.device)`. Ensures dtype consistency under LibTorch too.
3. **The failing `pair_matgl/kk (GPU, Kokkos)` check hit the 24 h timeout** rather than a test failure — likely a GPU runner queue/hang issue; re-run after the rebase before drawing conclusions.
4. The `python_reference.py` dual-convention printout (ASE σ vs. LAMMPS P) directly addresses how the sign defect survived review — good addition.

toohardtogetname:
Thanks for merging. Closing the loop on the 2026-09-01 review items, since none of them got a reply before the merge:

1. **Rebase onto #825** — the branch had already merged main after #825 (commit c403b03, "adopt #825's captured-local lambda"), so the `pair_matgl_kokkos.cpp` hunks were on top of the local-capture rewrite when this was merged; nothing further was needed.
2. **`_basis.py` dtype at construction** — done in #832: `torch.arange(max_n, dtype=matgl.float_th, device=r.device)` and `torch.tensor(1.0, dtype=matgl.float_th, device=r.device)`. Basis tests pass on CPU; a GPU single point through `PESCalculator` on a 1,248-atom cell matches main without the change to within the run-to-run spread of the same code (forces 3.0e-6 eV/A vs a 2.6-3.2e-6 floor; stress 1.3 bar vs 1.4-1.7 bar), energy identical.
3. **Kokkos GPU check** — it ended CANCELLED at the 24 h timeout again, so it never reported on this PR. For the record, the merged `_basis.py` device fix is exercised on GPU in the check above, and the Kokkos ghost-row trimming was verified on an H200 (hopper, same-build single-point parity against the CPU style and an independent implementation, see the PR body) before the merge. If the self-hosted runner comes back, a re-run on main would close that gap on your side.

PR #831, opened 2026-09-07
Fix bugs in building the Kokkos implementation of LAMMPS

mkphuthi:
# Fix ML-MATGL LAMMPS/Kokkos build, checkpoint export, and GPU basis bug

## Summary

Branch `mkphuthi/lammps-cmake-fixes`, based on `materialyzeai/matgl@dcf5fd3`
(2026-09-03). Fixes a chain of failures that kept `pair_style matgl` /
`matgl/kk` from working end-to-end on any machine with latest MatGL.

Major changes:

- **fix**: `cmake/ML-MATGL.cmake` never called `RegisterStyles()`, so
  `pair_style matgl` was silently never registered
- **fix**: the two `include(...)` lines were documented to go at the *end* of
  `cmake/CMakeLists.txt`, after `GenerateStyleHeaders()` already ran — too
  late for `RegisterStyles()` to have any effect.
- **feat**: `lammps/src/ML-MATGL/patch_lammps.py` — a small script that
  inserts the two includes in the correct place in a stock LAMMPS
  `cmake/CMakeLists.txt`. Deliberately simple: requires a clean LAMMPS tree,
  refuses if `CMakeLists.txt` already mentions `ML-MATGL`, no dry-run/backup
  machinery.
- **fix**: `pair_matgl.cpp`'s hardcoded a NERSC path
- **fix**: `export_matgl_checkpoint.py` imported `matgl.graph._compute_pyg`
  and `matgl.ext._lammps`, neither of which exists on current `matgl` main
  (`matgl.graph._compute`, `matgl.ext.lammps`). Rewritten to call
  `matgl.ext.lammps.export_lammps_model` directly, matching `mgl
  create-lammps-model`.
- **fix**: `pair_matgl`'s constructor now disables TorchScript's TensorExpr
  fuser (`torch::jit::setTensorExprFuserEnabled(false)`, overridable with
  `MATGL_TORCH_FUSER=1`). The fuser compiles fused CUDA kernels via NVRTC at
  run time, which needs a `libnvrtc-builtins` matching libtorch's exact CUDA
  minor version.
- **fix**: `matgl.layers._basis.spherical_bessel_smooth` built its basis-index
  tensor (`torch.arange(max_n)`) and a `dn` seed constant on the CPU
  regardless of the input's device. The TensorNet layer never hits this (it
  caches the equivalent constants as module buffers that move with `.to()`),
  but the LAMMPS export's `_TensorNetKernel` calls this functional form
  directly, so `pair_style matgl/kk` crashed with "Expected all tensors to be
  on the same device, cuda:0 and cpu" on the first GPU forward pass.
- **docs**: `lammps/README.md` build instructions rewritten.

## Numerical behaviour

`spherical_bessel_smooth`'s output values shift by floating-point-reassociation
noise only (max `total_energy_local` diff `6e-5` eV out of `~750` eV on a
144-atom NaCl cell, i.e. ~1e-7 relative) because building the same constants
via `dtype=r.dtype` instead of the ambient default dtype changes evaluation
order slightly. Not a physical change; existing test tolerances (`1e-5` /
`1e-4` / `1e-3` in `lammps/README.md`'s parity check) comfortably cover it.

## Checklist

- [x] Google format docstrings added (patch_lammps.py, export_matgl_checkpoint.py). Checked with `ruff`.
- [x] Type annotations included on new/changed Python. Checked with `mypy`.
- [x] Tests added for the basis-device fix (`tests/layers/test_basis.py`).
- [ ] `duecredit`/`@due.dcite` — not applicable (no new algorithm from a paper).


shyuep:
@drakeyu review asap.

shyuep:
Thanks. @mkphuthi there are some merge conflicts from an earlier PR. Probably fixing some of the same things. Can you fix?

mkphuthi:
@toohardtogetname Looks like you had pushed changes to fix the issues I highlight and they are slightly different in philosophy. I have opted to make the process simpler as 1) run patch_lammps.py to setup the source files 2) run cmake and build. This is as opposed to having users patch/edit/move files around manually. I also opted to not copy the matgl kokkos files into the lammps kokkos directory. Should I just overwrite with my implementation?

drakeyu:
Hi,

I've reviewed the PR. It updates the installation process so the package works outside NERSC, and includes a genuine GPU correctness bug fix in the basis layer. I checked out the branch in a clean worktree, compiled the executable, and ran a few simulations — everything works as expected.

shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff and CI only — no code executed locally, no human review).

Valuable PR — the `RegisterStyles()` omission and the include-ordering trap explain a class of "Unrecognized pair style matgl" reports, and removing the hardcoded NERSC paths is overdue. CPU CI is green. Items below, blocking first.

**Blocking**

1. **Branch is conflicting (`mergeable=CONFLICTING`).** Based on `dcf5fd3` (2026-09-03); `_basis.py` has moved since. Your hunk removes `torch.arange(max_n).type(...)` but main already carries `device=r.device` from #828, and #832 is open right now rewriting the same two lines as `dtype=matgl.float_th, device=r.device`. Rebase, and coordinate so one PR owns the hunk.

2. **`dtype=r.dtype` vs `matgl.float_th`.** Your `test_smooth_sbf_functional_follows_input_device_and_dtype` asserts float64 passthrough, which #832's approach breaks. I agree with your choice — a free function called directly by the TorchScript export should follow its input, not a module-level global — but it needs to be an explicit decision, not a merge-order accident. `torch.ones_like(en[0, 0])` for the `dn` seed is a nice touch: device-, dtype- and TorchScript-safe in one. The `meta`-device test standing in for CUDA is a good way to guard this without GPU CI.

3. **`pair_matgl/kk (GPU, Kokkos)` is still pending** — the one check that exercises what this PR fixes. Don't merge until it reports.

4. **Shell injection / quoting in `PairMATGL::coeff`.** `std::system("'" + python_exe + "' '" + export_script + "' '" + dir + "' ...")` breaks on any path containing a single quote, and `MATGL_PYTHON` / `MATGL_EXPORT_SCRIPT` / the `pair_coeff` directory are now all user-supplied, so a crafted value escapes the quoting and runs arbitrary commands. Pre-existing, but this PR makes all three inputs configurable. Prefer `posix_spawn`/`fork`+`execv` with an argv array, or at minimum reject arguments containing `'`.

**Non-blocking**

5. `<torch/csrc/jit/passes/tensorexpr_fuser.h>` is a libtorch internal header, not stable public API — it has moved between versions and will break the build silently across libtorch upgrades. Since the comment already notes `PYTORCH_TENSOREXPR=0` is the equivalent switch, consider setting that env var (or documenting it in the README) instead of pulling in `csrc`. Related: `setTensorExprFuserEnabled(false)` in the constructor is a process-global side effect that also disables fusion for any other libtorch pair style in the same run — worth a README note.

6. `ML-MATGL-KOKKOS.cmake` now calls `find_package(Torch REQUIRED)` and re-applies `target_link_libraries(lammps PRIVATE ${TORCH_LIBRARIES})` / appends `${TORCH_CXX_FLAGS}` after `ML-MATGL.cmake` already did both on the same target. Harmless in practice but the flags get appended twice; guard with a sentinel (`if(NOT DEFINED ML_MATGL_TORCH_CONFIGURED)`) so the layering is explicit. Also note the file header now says "include after the base snippet **and** set both PKG flags", while `patch_lammps.py` unconditionally inserts both includes — make sure a KOKKOS-off build still configures cleanly (the early `return()` should cover it, but the docs and the script now say slightly different things).

7. `patch_lammps.py`: `end = next(i for i in range(start, len(lines)) if lines[i].startswith("endforeach()"))` has no default, so a LAMMPS tree with an indented or reformatted `endforeach()` gives an uncaught `StopIteration` traceback rather than the clean `sys.exit(...)` message every other failure path produces. Add a default and an error string. The refuse-if-already-patched guard and the `GenerateStyleHeaders()` sanity check are well judged.

8. `export_matgl_checkpoint.py`: dropping the `create_line_graph_torch` shim in favour of `matgl.ext.lammps.export_lammps_model` is the right call. The new `print` reads `wrapper.r_max` / `wrapper.n_species` from the return value — worth a quick check that the current `export_lammps_model` signature returns the wrapper and exposes both attributes, since nothing in CI covers this script. `from __future__ import annotations` is a no-op here (no annotations in the module).

9. Consider splitting the `_basis.py` + test change out into its own PR (or letting #832 carry it). The LAMMPS build fixes and a `src/matgl` numerical change have very different review and release risk, and right now the `src/` change is the only part CI actually tests.

shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff and CI only — no code executed locally, no human review).

Re-review of `0d15e85a`. Blocker 1 from the previous review is resolved — the branch is now `MERGEABLE` and I confirmed zero conflict markers remain in all seven touched files. Thanks for the careful resolution note in the commit message; restoring the dropped `ML_MATGL_DIR` block was the right catch (every `target_sources()` below it referenced an undefined variable).

**Still blocking**

1. **`pair_matgl/kk (GPU, Kokkos)` has been `queued` for ~2 days** (job `102187287169`, run `34263582073`, queued 2026-09-08T18:31Z). Everything else on `0d15e85a` is green, but the self-hosted GPU runner looks offline or wedged, so the one check that exercises what this PR fixes has never executed. Either get the runner picked up or paste the manual `matgl/kk` parity numbers into the PR body; don't merge on CPU-green alone.

2. **Shell injection in `PairMATGL::coeff` is unchanged** (`pair_matgl.cpp:198-200`): `std::system("'" + python_exe + "' '" + export_script + "' '" + dir + "' ...")`. `MATGL_PYTHON`, `MATGL_EXPORT_SCRIPT` and the `pair_coeff` directory are all user-supplied, and a single quote in any of them escapes the quoting. `posix_spawn`/`fork`+`execv` with an argv array, or at minimum reject arguments containing `'`.

**New**

3. **The `_basis.py` dtype story doesn't quite match the test.** After the merge the hunk is one line, `dn = [torch.ones_like(en[0, 0])]`, and `en` derives from `n = torch.arange(max_n, device=r.device).type(dtype=matgl.float_th)` — so the constants track `matgl.float_th`, not `r.dtype`. `test_smooth_sbf_functional_follows_input_device_and_dtype`'s float64 assertion passes by type promotion against `r`, not because anything follows the input dtype. The device fix is real and `ones_like` is the right primitive (device-, dtype- and TorchScript-safe); please either rename the test or make the intent explicit in the docstring, since as written it will keep passing even if the constants regress to a fixed dtype.

4. **Overlap with #832 is now a direct conflict.** #832 rewrites the same two lines as `torch.arange(max_n, dtype=matgl.float_th, device=r.device)` / `torch.tensor(1.0, dtype=matgl.float_th, device=r.device)`. Whichever merges second will conflict, and #832's `dn` reverts your `ones_like`. Decide now which PR owns the hunk — my preference is #832 takes the `arange` cleanup and this PR keeps `ones_like`, but it should be a decision, not a merge-order outcome.

5. **Squash-merge this one.** `8c38ae96` committed literal conflict markers in `_basis.py` and `ML-MATGL.cmake`; `0d15e85a` cleans them up. A merge commit would put a non-building tree in main's first-parent history and break `git bisect`.

**Carried over, non-blocking (all unchanged in `0d15e85a`)**

6. `patch_lammps.py:58` — `next(i for i in range(start, len(lines)) if lines[i].startswith("endforeach()"))` still has no default, so a reformatted or indented `endforeach()` raises an uncaught `StopIteration` instead of the clean `sys.exit()` used everywhere else. Line 55 already does this correctly with `None` as the default; mirror it.

7. `ML-MATGL-KOKKOS.cmake:47,66-68` still re-runs `find_package(Torch REQUIRED)` and re-applies `target_link_libraries` / appends `TORCH_CXX_FLAGS` to the same target `ML-MATGL.cmake` already configured. Harmless but the flags land twice; a `ML_MATGL_TORCH_CONFIGURED` sentinel makes the layering explicit.

8. `pair_matgl.cpp:42` still pulls `<torch/csrc/jit/passes/tensorexpr_fuser.h>`, a libtorch internal header that has moved between releases. Setting `PYTORCH_TENSOREXPR=0` (documented in the README) avoids the future build break. Also worth a README note that `setTensorExprFuserEnabled(false)` is process-global and affects any other libtorch pair style in the same run.

9. `export_matgl_checkpoint.py` — dropping the unconditional top-level import makes the `ModuleNotFoundError` fallback reachable, which was the actual bug. Nothing in CI covers this script, so please confirm by hand that `export_lammps_model` returns a wrapper exposing both `r_max` and `n_species`.

toohardtogetname:
Heads-up on the `_basis.py` hunk: #832 now changes the line above yours, which makes your `dn` change redundant rather than conflicting.

#832 was originally `dtype=matgl.float_th` on both lines. Following review it now sets `n = torch.arange(max_n, dtype=r.dtype, device=r.device)`, so `en` and therefore `torch.ones_like(en[0, 0])` already follow `r`'s dtype without the second line being touched. If #832 lands first, your `dn = [torch.ones_like(en[0, 0])]` becomes a no-op you can drop; if yours lands first, #832 will carry it unchanged and the result is identical either way. Nothing needed from you right now, just so a conflict marker on that line does not come as a surprise.

Two measured details in case they are useful for your LAMMPS work, since the export path is what both of us care about here:

The `arange` line is the one that matters for the export. With `matgl.float_th = float32` and a float64 `r` (which is what the TorchScript `_TensorNetKernel` path hits), `n` stayed float32 while `r * n` promoted to float64, so the `dn` recursion ran below the precision of the rest of the expression. Loading the pre-change and post-change modules side by side in one process, on 257 distances with `max_n=9`: both library-consistent cases (`float_th` matching `r`) are bit-identical, and that mixed case goes from 2.757e-07 to exactly 0 against an all-float64 reference. Your `ones_like` alone does not close it, because `en` is still float32 there.

Your new `test_smooth_sbf_functional_follows_input_device_and_dtype` also passes against the *old* `matgl.float_th` version, 14/14 including the scripted parametrisation. `fn(r.double()).dtype == torch.float64` holds either way, because the output dtype comes from promotion rather than from the constants. Not a criticism of the test, which is the right test to have and is what caught this for me; just worth knowing it does not fail on the thing it is guarding against, so the dtype the recursion *runs in* is worth asserting too if you want it locked down.

One behaviour change to flag on #832 while you are looking at this hunk: with `matgl.float_th = torch.float64` and a float32 `r`, the function now returns float32 where it previously returned float64. Inside matgl that combination does not arise since `r` is built in `float_th`, but it is reachable from outside.

shyuep:
> 🤖 **Automated PR review** generated by Claude on behalf of @shyuep. Not a human review — treat as a first pass.

Clear improvement. Removing the hard-coded NERSC paths (`/global/cfs/cdirs/m4845/...`) from `pair_matgl.cpp` alone justifies this — that made the checkpoint-export path unusable off Perlmutter. Moving `RegisterStyles()` into the snippets with documented placement before `GenerateStyleHeaders()`, and automating the CMakeLists patch, removes the most error-prone part of the build instructions. Disabling the TensorExpr fuser is a good catch: the `libnvrtc-builtins` mismatch is a common HPC failure and the fused kernels buy little for a matmul/scatter-dominated GNN.

CI: 10/11 green, `pair_matgl/kk (GPU, Kokkos)` still pending.

## Issues

1. **`std::system` with hand-rolled quoting.** `"'" + python_exe + "' '" + export_script + "' '" + dir.string() + "' ..."` breaks on any path containing a single quote, and `dir` comes from user-supplied `pair_coeff` input. Prefer `fork`/`execv` (no shell) or, at minimum, reject paths containing `'` with a clear `error->one`. Low severity given the audience, but it is a trivially avoidable shell-injection surface.

2. **`file(GLOB ...)` for sources.** CMake does not re-glob on incremental builds, so adding a `.cpp` to `src/ML-MATGL` silently fails to compile until someone re-runs `cmake`. Use an explicit source list, or `CONFIGURE_DEPENDS`. The Kokkos snippet globs a single known filename, which is the same as naming it directly — just name it.

3. **`setTensorExprFuserEnabled(false)` is process-global state set in a per-instance constructor.** Constructing a second pair style, or any other libtorch consumer in the same process, inherits the change. Worth a comment noting it is intentionally global, and consider guarding with a `static std::once_flag` so the intent is explicit.

4. **`patch_lammps.py` error handling is inconsistent.** Most failures `sys.exit(...)` with a readable message, but `end = next(i for i in range(start, len(lines)) if lines[i].startswith("endforeach()"))` has no default and will surface as a bare `StopIteration` traceback. Add a default and exit the same way as its neighbours.

5. **The clean-tree requirement makes re-patching impossible.** Refusing when `ML-MATGL` already appears means a user who patched, upgraded matgl, and wants to re-point the include must hand-edit or `git checkout` first. A `--force` flag that replaces the existing block would be friendlier.

## Scope / correctness

6. **`tests/layers/test_basis.py` looks orphaned.** The new `test_smooth_sbf_functional_follows_input_device_and_dtype` says it "guards the fix", but no change to `spherical_bessel_smooth` (or anything under `src/`) is in this PR — all seven files are LAMMPS build plumbing plus this test. Either the fix landed elsewhere (say so, and link it) or it is missing here. Also unrelated to the stated Kokkos-build scope; consider splitting.

7. **`export_matgl_checkpoint.py` now calls `potential.eval()`** before export. Correct, and probably a real behaviour fix (dropout/batchnorm in train mode would have been baked into exported models) — but it is buried and unmentioned. Call it out in the PR description; anyone with a previously exported `lammps_model.pt` may want to re-export.

8. The README says the cached `lammps_model.pt` is only regenerated when `model.json`/`state.pt` is newer, and that users must delete it by hand after upgrading matgl. That is a footgun — a version stamp written alongside the cache would let the pair style invalidate automatically.

## Nits

- README step numbering: "patch_lammps.py (step 2 above)" — it is step 1 in the block above it.
- Both snippets call `find_package(Torch REQUIRED)` independently. Fine and intentional per the comment, but the duplicated `TORCH_LIBRARIES` emptiness check and `TORCH_CXX_FLAGS` block could live in a shared fragment.
- Mixed em-dash/`--` usage across the CMake comments after this PR; minor, but the Kokkos file now uses both.

PR #832, opened 2026-09-08
spherical_bessel_smooth: construct index and seed tensors with dtype (follow-up to #828)

toohardtogetname:
## Summary

Follow-up to the #828 review (item 2): `spherical_bessel_smooth` built its
index tensor as `torch.arange(max_n, device=r.device).type(dtype=matgl.float_th)`
(two allocations, dtype applied after the fact) and seeded the `dn`
recursion with `torch.tensor(1.0, device=r.device)` (default dtype, not
`matgl.float_th`). Both now carry `matgl.float_th` at construction:

```python
n = torch.arange(max_n, dtype=matgl.float_th, device=r.device)[None, :]
...
dn = [torch.tensor(1.0, dtype=matgl.float_th, device=r.device)]
```

No behavioural change is intended; the point is dtype consistency under
LibTorch/TorchScript and one allocation fewer per call.

## Tests

- `tests/layers/test_basis.py`: 13 passed (CPU, matgl main + this patch).
- GPU: TensorNet-PES-MatPES-r2SCAN-2025.2 single point on a 1,248-atom
  Li3PO4-TaCl5 cell through `PESCalculator` (energy, forces, stress), matgl
  main + this patch vs main without it (only these two lines differ): energy identical, max |dF| = 3.0e-6 eV/A, max |dS| = 1.3 bar -- inside the run-to-run spread of the same code on the same GPU (2.6-3.2e-6 eV/A, 1.4-1.7 bar; fp32 scatter ordering). H200, hopper.


shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff and CI only — no code executed locally, no human review).

Two-line dtype hygiene change, all 9 checks green. The mechanics are right: `torch.arange(max_n, dtype=..., device=...)` drops one allocation, and seeding `dn` with an explicitly typed scalar fixes a real latent bug — `torch.tensor(1.0)` picked up torch's ambient default dtype, so with `matgl.float_th == torch.float64` the `dn` recursion started in float32.

**One thing to settle before merge — `matgl.float_th` vs `r.dtype`.** #831 patches these exact two lines the other way: `dtype=r.dtype, device=r.device` and `dn = [torch.ones_like(en[0, 0])]`, plus a new `tests/layers/test_basis.py` case asserting `spherical_bessel_smooth(r.double(), 5.0, 3).dtype == torch.float64`. That assertion fails under this PR's approach, so the two PRs are in direct conflict over the same hunk and one of them should own it.

My read is that `r.dtype` is the better choice here:

- This functional form is what the TorchScript LAMMPS export (`_TensorNetKernel`) calls directly with whatever tensor it holds. Tying its constants to a module-level global rather than to the input is the wrong coupling for a free function — `matgl.float_th` is a sensible default for *module construction*, not for a function handed an arbitrary `Tensor`.
- With `matgl.float_th = float32` and a float64 `r`, you don't get a clean float32 path: `r * n` and the `_sinc` products promote back to float64 while `en`/`dn` stay float32, so the recursion runs at lower precision than the rest of the expression. `r.dtype` keeps the whole computation on one dtype.

Suggested resolution: rebase this onto #831 (or fold these two lines into #831 and close this), landing `dtype=r.dtype, device=r.device`. If you do prefer the global, #831's dtype-passthrough test needs to change with it.

Minor: the GPU parity numbers in the description (max |dF| = 3.0e-6 eV/Å inside run-to-run spread) are exactly the right evidence for a "no behavioural change" claim — worth putting in the commit message so it survives the squash.

toohardtogetname:
Thanks — agreed on `r.dtype`, and pushed (f72eeab). Two corrections to the premise, both measured rather than argued, because they change what has to move.

**#831 patches one of these lines, not two.** Its `_basis.py` hunk is only `dn = [torch.ones_like(en[0, 0])]`; the `torch.arange(...)` line is untouched, so `n` — and therefore `en`, and therefore `ones_like(en[0, 0])` — still comes from `matgl.float_th` on that branch. The two PRs were not in conflict over the same hunk.

**#831's new test passes under the `matgl.float_th` version too.** I applied `tests/layers/test_basis.py` from #831 on top of #832 as it stood and ran it: 14/14 green, both the `scripted` and eager parametrisations. `fn(r.double()).dtype == torch.float64` holds either way, because `r * n` promotes the output even when `n` is float32. What it does *not* catch is the thing the assertion is meant to protect, which is the point below.

**The precision argument is right, and only the `arange` line fixes it.** Loading the pre-PR module and the patched module side by side in one process and calling the real function on identical inputs (257 distances, `max_n=9`):

| `matgl.float_th` | `r` | before → after | max abs difference |
|---|---|---|---|
| float32 | float32 | float32 → float32 | 0 (bit-identical) |
| float64 | float64 | float64 → float64 | 0 (bit-identical) |
| float32 | float64 | float64 → float64 | error against an all-float64 reference falls from 2.757e-07 to exactly 0 |

The third row is the export path. `en`/`dn` stayed float32 while the rest of the expression promoted, so the `dn` recursion ran below the precision of everything around it — and `ones_like(en[0, 0])` alone does not fix that, since `en` is still float32 there. With the `arange` line moved as well, `dn` inherits `r`'s dtype and the row goes to zero. That also makes #831's `dn` hunk redundant rather than conflicting: same value, arrived at from the input instead of the global.

**One behaviour change worth naming before merge.** The coupling is now to the argument in both directions, so with `matgl.float_th = torch.float64` and a float32 `r`, the function returns float32 where it previously returned float64. Inside matgl that combination does not arise — `r` is built in `float_th` — so no in-library path changes, and both bit-identical rows above are exactly that statement. It is reachable from outside, though, and it is the price of the free function following its argument.

97 tests in `tests/layers` pass on the pushed commit, including #831's new one.

On the minor point: the GPU parity numbers are now in the commit message.

shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff and CI only — no code executed).

Re-review after f72eeab. Both lines are right now. `en` derives from `n`, so it already carries `r.dtype` and `r.device`, which makes `torch.ones_like(en[0, 0])` the correct seed for `dn` — it drops the hardcoded `1.0` while matching dtype *and* device, where the previous `torch.tensor(1.0, device=r.device)` only matched device and silently pinned float32. Your correction on `r.dtype` over `matgl.float_th` stands.

One thing worth stating explicitly in the module, since this PR settles it de facto: the basis now follows **the input tensor's** dtype rather than the global `matgl.float_th`. Those coincide today because graph construction emits edge distances in `matgl.float_th`, so this is a no-op in practice. But it does move the failure mode: a caller who sets `matgl.float_th = torch.float64` and passes a float32 `r` previously got a float64 basis (compatible with float64 weights); they now get a float32 basis and the mismatch surfaces at the first matmul downstream, further from the cause. Following the input is the better contract for a pure function — just worth one line of comment so the next dtype question does not get re-litigated.

CI is fully green (lint, all six test legs, `test_grace_parity`, coverage) and the branch is mergeable-clean. LGTM.

PR #833, opened 2026-09-14
Require lightning >= 2.6.6 (CVE-2026-58659)

toohardtogetname:
## The problem

`lightning<=2.6.1` has no lower bound, so a resolver picks a version inside the affected range of [GHSA-qqmf-gpg7-g8gw / CVE-2026-58659](https://github.com/advisories/GHSA-qqmf-gpg7-g8gw) (high). `LightningModule.load_from_checkpoint` imports and calls whatever module path a checkpoint names in its `_instantiator` hyperparameter, which bypasses `weights_only=True`, so a crafted checkpoint runs arbitrary code on load. Lightning 2.6.6 adds an allowlist for `_instantiator` plus a subclass check on `_class_path`.

The advisory's machine-readable "first patched version" reads `2022.6.15`, which is an artefact of the package's old calendar versioning and makes the range render as `< 2022.6.15`. The advisory text is the accurate one: affected "through 2.6.5".

## How we ran into it

We hit this in **boltmd**, which depends on matgl. boltmd does not import lightning anywhere, and nothing on our side calls `load_from_checkpoint`, so no vulnerable code path is reachable for us. But lightning comes in transitively through matgl, matgl's cap holds it at 2.6.1, and the advisory therefore lands on our lockfile with no way to resolve it downstream. We are carrying a `[tool.uv] override-dependencies = ["lightning>=2.6.6"]` to get past it, which is the sort of workaround that is better fixed at the source.

Anyone else pinning matgl is in the same position, and unlike us some of them will actually load third-party checkpoints.

## The change

    -    "lightning<=2.6.1",
    +    "lightning>=2.6.6,<2.7",

The cap keeps the shape it has had since b392449, now pinned to the 2.6 series so patch releases come in without another manual bump, and a lower bound is added so the vulnerable versions cannot be selected.

## Verification

Full suite against lightning 2.6.6, on Python 3.12 with matgl 4.0.3:

- **428 tests collected, 0 failures.** The three skips are `jax`, `nvalchemi-toolkit` and `tensorflow`, all optional dependencies that are absent in that environment and none of them lightning-related.
- `tests/utils`, the lightning-facing ones, pass on their own as well.
- `PotentialLightningModule` builds, `pl.Trainer` builds, `matgl.load_model` works.

The cap looks incidental rather than protective: it entered as `<=2.4.0` in b392449 ("Tru to fix requirements") and has been bumped by hand since. If you would rather keep a hard ceiling at the current release, `>=2.6.6,<=2.6.6` also solves the advisory, but it will need bumping again on the next patch.


shyuep:
Is the upper bound even needed?

toohardtogetname:
No, and the history says so. Dropped it in cb71dfc, so the line is now just `lightning>=2.6.6`.

Every change to that constraint since it appeared has been a mechanical Dependabot bump to whatever was current, never a response to a break:

| commit | date | value |
|---|---|---|
| b392449 | 2024-08-13 | `<=2.4.0` (first cap, "Tru to fix requirements") |
| ... | | seven further Dependabot bumps |
| 537fcb3 | 2025-10-30 | `<=2.6.0.dev20251026` |
| 284c83e | 2025-11-03 | `<=2.6.0.dev20251102` |
| 40d83ba | 2025-11-14 | `<=2.6.0.dev20251109` |
| e9d4354 | 2025-11-19 | `<=2.6.0.dev20251116` |
| 68a7243 | 2025-11-24 | `<=2.6.0.dev20251123` |
| d1a67f7 | 2025-11-29 | `<=2.6.0` |
| 31fd7ce | 2026-02-25 | `<=2.6.1` |

Four of those pin a `.dev` snapshot, which is a cap chasing nightlies rather than guarding against anything. The failure mode is visible in the last row: the bumps stopped in February, 2.6.4, 2.6.5 and 2.6.6 shipped, and the requirement sat three releases behind on the one carrying the advisory. A cap that is maintained by a bot tracks the latest release with a lag, and the lag is the whole problem.

The surface being protected is also small. matgl touches lightning through two files, `utils/training.py` and `utils/callbacks.py`, and four symbols: `pl.Trainer` (14 uses), `pl.LightningModule` (9), `pl.seed_everything` (2), `pl.Callback` (1). That is lightning's most stable public API. CI resolves lightning from this file with no constraint of its own, so a real break in some future release surfaces as a failing test run rather than silently, which is the case where a cap would have earned its keep.

The verification from the first commit stands unchanged, since dropping an upper bound does not move today's resolved version: 428 tests pass against 2.6.6 on Python 3.12, `tests/utils` included, with skips only for the absent jax, alchmtk and tensorflow extras.

Happy to put `<2.7` back if you would rather keep a ceiling; it solves the advisory either way.

